PR #20, opened 2023-05-07
improve display on mobile phones

domel:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/20.html" title="Last updated on May 17, 2023, 6:41 PM UTC (e911b83)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/20/9746bf7...e911b83.html" title="Last updated on May 17, 2023, 6:41 PM UTC (e911b83)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -65,8 +65,13 @@
     .grammar_comment { color: #A52A2A; font-style: italic; }
     .hl-bold { font-weight: bold; color: #0a3; }
     code { color: #ff4500; }  /* Old W3C Style */
-    table { word-break: break-all; max-width: 80vw; }
-    table.ex, .ex th, .ex td { border: none; padding: 0; }
+    @media (max-width: 767px) {
+      table { word-break: break-all; max-width: 80vw; }
+      .separated thead tr th { border:1px solid black; padding: .2em; min-width: 10vw }
+      .separated tbody tr td { border:1px solid black; text-align: center; min-width: 10vw }
+      table { word-break: break-all; max-width: 80vw; }
+      table.ex, .ex th, .ex td { border: none; padding: 0; }
+    }
     .ex th { text-align: center; }
   </style>
 </head>
--- a/spec/index.html
+++ b/spec/index.html
@@ -65,13 +65,19 @@
     .grammar_comment { color: #A52A2A; font-style: italic; }
     .hl-bold { font-weight: bold; color: #0a3; }
     code { color: #ff4500; }  /* Old W3C Style */
+    @media (max-width: 850px) {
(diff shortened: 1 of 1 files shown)


gkellogg (approved):

gkellogg (commented):
It's odd that the tables are full-width for most widths, but then narrower for small sizes. It's in the small sizes where you would want to use all the available width. Can't see the effect on the grammar itself, because of Preview limitations, but I presume it's the same.

pfps:
@gkellogg See also https://github.com/w3c/rdf-semantics/pull/30

TallTed:
In the **Context where each kind of escape sequence can be used**, the contrast between the dark-green background and the black text thereon is problematic. A relatively simple fix would be to use white text for these; alternatively, the background could be made a lighter green.

domel:
@TallTed done

TallTed:
The table in [§2.5.2 Numbers](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/20.html#abbrev) would benefit greatly by narrowing the first 2 columns and passing that width to the 4th column. (The 3rd column is only slightly wider than its widest value, so should remain very close to the same width.)

Similarly, the 2nd column of the table in [§7.2 RDF Term Constructors](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/20.html#sec-parsing-terms) could pass some width to the 3rd.

In both cases, words are broken to wrap and keep the table close to if not smaller than the viewport. Better not to break words, and to overflow the viewport and need soome horizontal scrolling.

domel:
@TallTed I'm not happy with hacking tables like this. IMO it should be decided by User Agent, but OK I hacked those tables you mentioned. Hope you like it now.

pfps:
This PR has the effect of making the font size smaller for tables when the window with becomes small (but still over 1/2 of the width of my laptop's screen with what I consider to be a reasonable font size).   This has the decidedly negative effect of making the tables harder to read.   There does not appear to be any benefit to this for many, and probably all, of the tables as they could be rendered at the normal font size with no problems.

Here is an example:
Screenshot from 2023-05-18 21-00-05

TallTed:
Please be aware that I will be offline and unresponsive through May 30. That will not change my interest in getting this resolved in the best way for all, which it is not yet. I would have suggested not merging this PR, given the ongoing mixed feelings about its effect. I will suggest that there should be a new issue opened on this topic for the Turtle document, regardless of how the others stand.

pfps:
See https://github.com/w3c/rdf-star-wg/issues/61

PR #32, opened 2023-06-12
Literal escape language

gkellogg:
* Update EBNF with improved character escapes.
* Minor content improvements.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/32.html" title="Last updated on Jun 15, 2023, 8:57 PM UTC (ac34f26)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/32/158031a...ac34f26.html" title="Last updated on Jun 15, 2023, 8:57 PM UTC (ac34f26)">Diff</a>


spec/turtle-bnf.html
```diff
@@ -234,15 +233,15 @@ <h3 id="terminals">Productions for terminals</h3>
...
+      <td>'<code class="grammar-literal">&quot;</code>' <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">^</code><code class="grammar-char-escape"><abbr title="dquote">#x22</abbr></code><code class="grammar-char-escape"><abbr title="back slash">#x5C</abbr></code><code class="grammar-char-escape"><abbr title="new line">#x0A</abbr></code><code class="grammar-char-escape"><abbr title="carriage return">#x0D</abbr></code><code class="grammar-brac">]</code> <code class="grammar-alt">|</code> <a href="#grammar-production-ECHAR">ECHAR</a> <code class="grammar-alt">|</code> <a href="#grammar-production-UCHAR">UCHAR</a><code class="grammar-paren">)</code><code class="grammar-star">*</code> '<code class="grammar-literal">&quot;</code>'
```
afs, line 236: "backslash" seems to be is one word (web search ; wikipedia). In unicode it is called "reverse solidus" but I think most people would find "backslash" clearer.

gkellogg: In N-Quads, @TallTed suggested "solidus"/"reverse solidus" which is certainly more _standardy_. I think the "back slash" came from auto-correct of "backslash".

Let's pick one or the other convention: "slash/backslash" or "solidus/reverse solidus". But, looking at other docs, it seems that we do use either "BACKSLASH" or "backslash". and I see no other use of "solidus".

TallTed: The trouble with `slash`/`backslash` is that people VERY often confuse the two, including when they're trying to explain which is which. This should be lessened by the inclusion of the actual `/`/`\` character, but I expect confusion to persist.

`solidus`/`reverse solidus` are not as popularly used, but I have not encountered any confusion between them.

I'll live with the group's decision either way.

afs: [SPARQL uses backslash](https://www.w3.org/TR/sparql11-query/) - two places.

PR #34, opened 2023-07-20
remove errant space character

TallTed:
***
<a href="https://pr-preview.s3.amazonaws.com/TallTed/rdf-turtle/pull/34.html" title="Last updated on Jul 20, 2023, 8:06 PM UTC (8e1470e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/34/fe482e6...TallTed:8e1470e.html" title="Last updated on Jul 20, 2023, 8:06 PM UTC (8e1470e)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -449,7 +449,7 @@
         (<span class="codepoint">U+0040</span>).
         If there is no language tag, there may be a <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>,
         preceeded by '<code>^^</code>'
-        (<span class="codepoint">U+005E</span> <span class="codepoint">U+005E</span>).
+        (<span class="codepoint">U+005E</span><span class="codepoint">U+005E</span>).
         The datatype IRI in Turtle may be written using either an
         <a href="#sec-iri">absolute IRI</a>,
         a <a href="#relative-iri">relative IRI reference</a>,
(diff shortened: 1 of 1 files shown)


gkellogg (approved):

afs:
In the preview, it becomes `(U+005EU+005E)`.
Is that intended?

(the diff is broken and showing various things that do not agree with the current doc or the preview)

TallTed:
Yes, the intent is to remove the suggestion that there's a space between, by removing the space between.  There might be some CSS tweak to be made to the `codepoint` class, to make it clearer that these are two codepoints...

Alternatively, the text might be tweaked to describe this as *two* of the singular `U+005E` character, only displayed and identified here as a singlet, until the complete example where there are two `^` between the URI and the lang.

domel (approved):

domel (approved):

afs:
Yes - saying the character `^` is `U+005E` would work.

gkellogg:
> Yes, the intent is to remove the suggestion that there's a space between, by removing the space between. There might be some CSS tweak to be made to the `codepoint` class, to make it clearer that these are two codepoints...

We could add a class to one or the other condpoints to add a margin after/before, but to a reader that may just look like a space. In fact, the `^^` are unambiguous as is, as they are UNICODE characters that represent themselves. It's only to someone reading a printed document, where the actual character is not apparent where there is ambiguity.

> Alternatively, the text might be tweaked to describe this as _two_ of the singular `U+005E` character, only displayed and identified here as a singlet, until the complete example where there are two `^` between the URI and the lang.

Probably better to describe it narratively as:

> (using two concatenated `^` characters each having the code point <span class="codepoint">U+005E</span>).

gkellogg (approved):

PR #46, opened 2023-09-30
Use table to make grammar literal terminals explicit

### LINKED ISSUE #45: Simplify character token representation
As suggested in https://github.com/w3c/rdf-turtle/pull/39#issuecomment-1738514256.

* Gather explicit token representations with code points in a "How to Read this Document" section.
* Change existing explicit code point annotations with references into the "How to Read this Document" section.
* Use tooltip (attribute `title=""`) on tokens to make it clear which Unicode character is used.
* Add a sentence that the grammar and tokens are expressed using the ASCII range of Unicode.

gkellogg:
… with simplified use within the document.

This creates a "Typographical Conventions" section with some general conventions, plus a table of Unicode characters used in the document and some sequences which are referenced from within the document using a simpler representation.

Fixes #45.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/46.html" title="Last updated on Oct 3, 2023, 7:19 PM UTC (2c0d68f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/46/e882b8a...2c0d68f.html" title="Last updated on Oct 3, 2023, 7:19 PM UTC (2c0d68f)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -197,214 +197,6 @@
     <a href="#sec-grammar" class="sectionRef">Turtle Grammar</a>
     and <a href="#sec-parsing" class="sectionRef">Parsing</a>.
   </p>
-
-<section id="typographical-conventions" class="informative">
-  <h3>Typographical conventions</h3>
-
-  <p>The following typographic conventions are used in this specification:</p>
-
-  <dl class="typography">
-    <dt><code>markup</code></dt><dd>
-      Markup (elements, attributes, properties),
-      machine processable values (string, characters, media types),
-      property name,
-      or a file name is in red-orange monospace font.</dd>
-    <dt><dfn data-no-xref="">definition</dfn></dt><dd>
-      A definition of a term, to be used elsewhere in this or other specifications,
-      is in bold and italics.</dd>
-    <dt><a data-lt="definition">definition reference</a></dt><dd>
-      A reference to a definition <em>in this document</em>
-      is underlined and is also an active link to the definition itself. </dd>
-    <dt><a href="">hyperlink</a></dt><dd>
-      A hyperlink is underlined and in blue.</dd>
-    <dt>[<a href="">reference</a>]</dt><dd>
-      A document reference (normative or informative) is enclosed in square brackets
-      and links to the references section.</dd>
-    <dt><a href="" class="cp-terminal"><code>:</code></a></dt><dd>
-      A grammar literal terminal with specific Unicode code points.
-      Grammar literal terminals used in this specification include the following:
-
-      <!-- Selected unicode character descriptions, taken from https://en.wikipedia.org/wiki/List_of_Unicode_characters -->
-      <table class="cp-definitions">
-        <thead>
-          <tr><th>Code</th><th>Glyph</th><th>Description</th></tr>
-        </thead>
-        <tbody>
-          <tr id="cp-tab">
-            <td><code class="codepoint">U+0009</code></td>
-            <td><code title="horizontal tab">HT</code></td>
-            <td>Horizontal tab</td>
-          </tr>
-          <tr id="cp-line-feed">
-            <td><code class="codepoint">U+000A</code></td>
-            <td><code title="line feed">LF</code></td>
-            <td>Line feed</td>
-          </tr>
-          <tr id="cp-carriage-return">
-            <td><code class="codepoint">U+000D</code></td>
-            <td><code title="carriage return">CR</code></td>
-            <td>Carriage return</td>
-          </tr>
-          <tr id="cp-quotation-mark">
-            <td><code class="codepoint">U+0022</code></td>
-            <td><code title="quotation mark">&quot;</code></td>
-            <td>Quotation mark</td>
-          </tr>
-          <tr id="cp-number-sign">
-            <td><code class="codepoint">U+0023</code></td>
-            <td><code title="number sign">#</code></td>
-            <td>Number sign</td>
-          </tr>
-          <tr id="cp-percent-sign">
-            <td><code class="codepoint">U+0025</code></td>
-            <td><code title="percent sign">%</code></td>
-            <td>Percent sign</td>
-          </tr>
-          <tr id="cp-apostrophe">
-            <td><code class="codepoint">U+0027</code></td>
-            <td><code title="apostrophe">'</code></td>
-            <td>Apostrophe</td>
-          </tr>
-          <tr id="cp-paren-left">
-            <td><code class="codepoint">U+0028</code></td>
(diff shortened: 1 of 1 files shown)


domel (approved):

domel:
Will a "Typographical Conventions" section also be available in TriG, N-Triples, N-Quads, etc.?

gkellogg:
If t hi is direction is agreeable, I’ll update the Editor’s Guide and do the same in other RDF syntax documents, where appropriate.

spec/index.html
```diff
@@ -191,22 +197,230 @@ <h2>Introduction</h2>
...
+      is in bold and italics.</dd>
+    <dt><a data-lt="definition">definition reference</a></dt><dd>
+      A reference to a definition <em>in this document</em>
+      is underlined and is also an active link to the definition itself. </dd>
+    <dt><a href="">hyperlink</a></dt><dd>
+      A hyperlink is underlined and in blue.</dd>
+    <dt>[<a href="">reference</a>]</dt><dd>
+      A document reference (normative or informative) is enclosed in square brackets
+      and links to the references section.</dd>
+    <dt><a href="" class="cp-terminal"><code>:</code></a></dt><dd>
```
afs, line 223: This is a floating `:`. See preview.

spec/index.html
```diff
@@ -191,22 +197,230 @@ <h2>Introduction</h2>
...
+      A grammar literal terminal with specific Unicode code points.
+      Grammar literal terminals used in this specification include the following:
+
+      <!-- Selected unicode character descriptions, taken from https://en.wikipedia.org/wiki/List_of_Unicode_characters -->
+      <table class="cp-definitions">
```
afs, line 228: This table is out of proportion for the introduction and just before the non-normative descriptive section. It just slows the first time reader down.

Here, it mentions `e` for example, leaving the query why not `a`, `b`, ...

(table = "A grammar literal terminal ..." to end `|}`)

At a minimum - a multiple column table.
Better, is to put the table at the end of the grammar 6.6.

If a mention is needed (and I don't think it is) with something like "A list of terminal mentioned in the grammar is <here>"

Down text still has: `U+`

>The characters [-](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/46.html#cp-hyphen), [‿](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/46.html#cp-undertie), [⁀](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/46.html#cp-character-tie), and the combining diacritical marks (U+0300 to U+036F)

afs (changes requested):

spec/index.html
gkellogg: As with the previous examples, this is intended to show how a code point reference is rendered. Of course, it could use a different character, or something else entirely. But, if we move the table down to the grammar section, the other typographic examples might not be worth carrying through.

spec/index.html
gkellogg: > This table is out of proportion for the introduction and just before the non-normative descriptive section. It just slows the first time reader down.

Yes, I thought about this, and your suggestion (below) to add it to the grammar section makes sense to me.

> Here, it mentions `e` for example, leaving the query why not `a`, `b`, ...

That's because `e` (and `E` are explicitly referenced (similar to `0` and `9`). All Unicode characters could be listed, but only a relatively small subset are actually referenced.

> (table = "`A grammar literal terminal ..." to end `|}`)

I don't understand this comment.

> At a minimum - a multiple column table. Better, is to put the table at the end of the grammar 6.6.

👍 

> If a mention is needed (and I don't think it is) with something like "A list of terminal mentioned in the grammar is "
> 
> Down text still has: `U+`
> 
> > The characters [-](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/46.html#cp-hyphen), [‿](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/46.html#cp-undertie), [⁀](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/46.html#cp-character-tie), and the combining diacritical marks (U+0300 to U+036F)

The diacritical marks don't have an associated Glyph, so were left as bare code points.

spec/index.html
```diff
@@ -94,7 +94,13 @@
 
     }
     .ex th { text-align: center; }
-  </style>
+
+    table.cp-definitions { border-collapse:collapse; background-color: #DDDDFF}
+
+    table.cp-definitions,
+    table.cp-definitions th,
+    table.cp-definitions td { border:1px solid black; padding:0.2em; }
```
TallTed, line 102: I don't see a quick way to do this selectively ... but I'd like the Glyph column to be center-aligned, while the others do make sense left-aligned.

spec/index.html
```diff
@@ -1293,6 +1295,180 @@ <h3>Grammar</h3>
...
+
+  <section>
+    <h2>Selected Terminal Literal Strings</h2>
+    <p>This document uses some specific terminal literal strings [[EBNF-NOTATION]].
+      To clarify the Unicode code points used for these terminal literal strings, the following
+      table describes specific characters and sequences used throughout this document.</p>
+
+    <!-- Selected unicode character descriptions, taken from https://en.wikipedia.org/wiki/List_of_Unicode_characters -->
+    <table class="cp-definitions">
+      <thead>
```
TallTed, line 1307: I suggest splitting the current table into 2 or maybe 3 sections (depends how many fit into the current body width) that are presented in 2 or 3 columns beneath the intro paragraph.

TallTed (commented):
Consider this an approval, whether my suggestions are taken or not.

spec/index.html
afs: I wasn't suggesting removing the rest of the typography section. (I'm neutral as to whether to have it or not.)

afs (approved):

spec/index.html
gkellogg: Although it may be generally useful, the motivation for putting it in was to have a place for the table, which isn't needed anymore. We can add something like the original back, although should probably do it universally. I personally don't think it's necessary.

spec/index.html
gkellogg: That would need to be a responsive layout do deal with screen-size issues. I think it would be great, but maybe should be addressed along with other responsive UI considerations.

spec/index.html
gkellogg: Done in https://github.com/w3c/rdf-turtle/pull/46/commits/2c0d68f7ae4336eaa95610e77d412be29ba8908e.

PR #51, opened 2023-12-20
Grammar updates for triple terms and occurrences.

gkellogg:
* Changes `quotedTriple` (and related) to `tripleOccurrence`.
* Adds `tripleTerm`.
* Changes `annotation` to allow an identifier. (Note, this change makes the grammar no longer context free).

### DIFF
--- a/spec/turtle.bnf
+++ b/spec/turtle.bnf
@@ -9,9 +9,9 @@ triples               ::= subject predicateObjectList | blankNodePropertyList pr
 predicateObjectList   ::= verb objectList (';' (verb objectList)? )*
 objectList            ::= object annotation? ( ',' object annotation? )*
 verb                  ::= predicate | 'a'
-subject               ::= iri | BlankNode | collection | quotedTriple
+subject               ::= iri | BlankNode | collection | tripleTerm | tripleOccurrence
 predicate             ::= iri
-object                ::= iri | BlankNode | collection | blankNodePropertyList | literal | quotedTriple
+object                ::= iri | BlankNode | collection | blankNodePropertyList | literal | tripleTerm | tripleOccurrence
 literal               ::= RDFLiteral | NumericLiteral | BooleanLiteral
 blankNodePropertyList ::= '[' predicateObjectList ']'
 collection            ::= '(' object* ')'
@@ -23,10 +23,11 @@ String                ::= STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE
 iri                   ::= IRIREF | PrefixedName
 PrefixedName          ::= PNAME_LN | PNAME_NS
 BlankNode             ::= BLANK_NODE_LABEL | ANON
-quotedTriple          ::= '<<' qtSubject predicate qtObject '>>'
-qtSubject             ::=	iri | BlankNode | quotedTriple
-qtObject              ::=	iri | BlankNode | literal | quotedTriple
-annotation            ::= '{|' predicateObjectList '|}'
+tripleTerm            ::= '<<(' ttSubject predicate ttObject ')>>'
+tripleOccurrence      ::= '<<' ( '|' (iri | BlankNode) '|' )? ttSubject predicate ttObject '>>'
+ttSubject             ::=	iri | BlankNode | tripleTerm | tripleOccurrence
+ttObject              ::=	iri | BlankNode | literal | tripleTerm | tripleOccurrence
+annotation            ::= '{|' ( (iri | BlankNode) '|' )? predicateObjectList '|}'
 
 @terminals
 
--- a/spec/turtle-bnf.html
+++ b/spec/turtle-bnf.html
@@ -1,4 +1,4 @@
-<!-- Generated with ebnf version 2.3.4. See https://github.com/dryruby/ebnf. -->
+<!-- Generated with ebnf version 2.5.0. See https://github.com/dryruby/ebnf. -->
 <table class="grammar">
   <tbody id="grammar-productions" class="ebnf">
     <tr id="grammar-production-turtleDoc">
@@ -71,7 +71,7 @@
       <td>[12]</td>
       <td><code>subject</code></td>
       <td>::=</td>
-      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-collection">collection</a> <code class="grammar-alt">|</code> <a href="#grammar-production-quotedTriple">quotedTriple</a></td>
+      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-collection">collection</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleTerm">tripleTerm</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleOccurrence">tripleOccurrence</a></td>
     </tr>
     <tr id="grammar-production-predicate">
       <td>[13]</td>
@@ -83,7 +83,7 @@
       <td>[14]</td>
       <td><code>object</code></td>
       <td>::=</td>
-      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-collection">collection</a> <code class="grammar-alt">|</code> <a href="#grammar-production-blankNodePropertyList">blankNodePropertyList</a> <code class="grammar-alt">|</code> <a href="#grammar-production-literal">literal</a> <code class="grammar-alt">|</code> <a href="#grammar-production-quotedTriple">quotedTriple</a></td>
+      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-collection">collection</a> <code class="grammar-alt">|</code> <a href="#grammar-production-blankNodePropertyList">blankNodePropertyList</a> <code class="grammar-alt">|</code> <a href="#grammar-production-literal">literal</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleTerm">tripleTerm</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleOccurrence">tripleOccurrence</a></td>
     </tr>
     <tr id="grammar-production-literal">
       <td>[15]</td>
@@ -145,29 +145,35 @@
       <td>::=</td>
       <td><a href="#grammar-production-BLANK_NODE_LABEL">BLANK_NODE_LABEL</a> <code class="grammar-alt">|</code> <a href="#grammar-production-ANON">ANON</a></td>
     </tr>
-    <tr id="grammar-production-quotedTriple">
+    <tr id="grammar-production-tripleTerm">
       <td>[25]</td>
-      <td><code>quotedTriple</code></td>
+      <td><code>tripleTerm</code></td>
       <td>::=</td>
-      <td>'<code class="grammar-literal">&lt;&lt;</code>' <a href="#grammar-production-qtSubject">qtSubject</a> <a href="#grammar-production-predicate">predicate</a> <a href="#grammar-production-qtObject">qtObject</a> '<code class="grammar-literal">&gt;&gt;</code>'</td>
+      <td>'<code class="grammar-literal">&lt;&lt;(</code>' <a href="#grammar-production-ttSubject">ttSubject</a> <a href="#grammar-production-predicate">predicate</a> <a href="#grammar-production-ttObject">ttObject</a> '<code class="grammar-literal">)&gt;&gt;</code>'</td>
     </tr>
-    <tr id="grammar-production-qtSubject">
+    <tr id="grammar-production-tripleOccurrence">
       <td>[26]</td>
-      <td><code>qtSubject</code></td>
+      <td><code>tripleOccurrence</code></td>
       <td>::=</td>
-      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-quotedTriple">quotedTriple</a></td>
+      <td>'<code class="grammar-literal">&lt;&lt;</code>' <code class="grammar-paren">(</code>'<code class="grammar-literal">|</code>' <code class="grammar-paren">(</code><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a><code class="grammar-paren">)</code> '<code class="grammar-literal">|</code>'<code class="grammar-paren">)</code><code class="grammar-opt">?</code> <a href="#grammar-production-ttSubject">ttSubject</a> <a href="#grammar-production-predicate">predicate</a> <a href="#grammar-production-ttObject">ttObject</a> '<code class="grammar-literal">&gt;&gt;</code>'</td>
     </tr>
-    <tr id="grammar-production-qtObject">
+    <tr id="grammar-production-ttSubject">
       <td>[27]</td>
-      <td><code>qtObject</code></td>
+      <td><code>ttSubject</code></td>
       <td>::=</td>
-      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-literal">literal</a> <code class="grammar-alt">|</code> <a href="#grammar-production-quotedTriple">quotedTriple</a></td>
+      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleTerm">tripleTerm</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleOccurrence">tripleOccurrence</a></td>
+    </tr>
+    <tr id="grammar-production-ttObject">
+      <td>[28]</td>
+      <td><code>ttObject</code></td>
+      <td>::=</td>
+      <td><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a> <code class="grammar-alt">|</code> <a href="#grammar-production-literal">literal</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleTerm">tripleTerm</a> <code class="grammar-alt">|</code> <a href="#grammar-production-tripleOccurrence">tripleOccurrence</a></td>
     </tr>
     <tr id="grammar-production-annotation">
-      <td>[28]</td>
+      <td>[29]</td>
       <td><code>annotation</code></td>
       <td>::=</td>
-      <td>'<code class="grammar-literal">{|</code>' <a href="#grammar-production-predicateObjectList">predicateObjectList</a> '<code class="grammar-literal">|}</code>'</td>
+      <td>'<code class="grammar-literal">{|</code>' <code class="grammar-paren">(</code><code class="grammar-paren">(</code><a href="#grammar-production-iri">iri</a> <code class="grammar-alt">|</code> <a href="#grammar-production-BlankNode">BlankNode</a><code class="grammar-paren">)</code> '<code class="grammar-literal">|</code>'<code class="grammar-paren">)</code><code class="grammar-opt">?</code> <a href="#grammar-production-predicateObjectList">predicateObjectList</a> '<code class="grammar-literal">|}</code>'</td>
     </tr>
     <tr id="grammar-declaration-terminals">
       <td colspan="4">
@@ -175,62 +181,62 @@
       </td>
     </tr>
     <tr id="grammar-production-IRIREF">
-      <td>[30]</td>
+      <td>[31]</td>
       <td><code>IRIREF</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">&lt;</code>' <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">^</code><code class="grammar-char-escape"><abbr title="null">#x00</abbr></code><code class="grammar-literal">-</code><code class="grammar-char-escape"><abbr title="space">#x20</abbr></code><code class="grammar-literal">&lt;&gt;&quot;{}|^`\</code><code class="grammar-brac">]</code> <code class="grammar-alt">|</code> <a href="#grammar-production-UCHAR">UCHAR</a><code class="grammar-paren">)</code><code class="grammar-star">*</code> '<code class="grammar-literal">&gt;</code>'
         <br/><span class="grammar_comment">/* #x00=NULL #x01-#x1F=control codes #x20=space */</span></td>
     </tr>
     <tr id="grammar-production-PNAME_NS">
-      <td>[31]</td>
+      <td>[32]</td>
       <td><code>PNAME_NS</code></td>
       <td>::=</td>
       <td><a href="#grammar-production-PN_PREFIX">PN_PREFIX</a><code class="grammar-opt">?</code> '<code class="grammar-literal">:</code>'</td>
     </tr>
     <tr id="grammar-production-PNAME_LN">
-      <td>[32]</td>
+      <td>[33]</td>
       <td><code>PNAME_LN</code></td>
       <td>::=</td>
       <td><a href="#grammar-production-PNAME_NS">PNAME_NS</a> <a href="#grammar-production-PN_LOCAL">PN_LOCAL</a></td>
     </tr>
     <tr id="grammar-production-BLANK_NODE_LABEL">
-      <td>[33]</td>
+      <td>[34]</td>
       <td><code>BLANK_NODE_LABEL</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">_:</code>' <code class="grammar-paren">(</code><a href="#grammar-production-PN_CHARS_U">PN_CHARS_U</a> <code class="grammar-alt">|</code> <code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-paren">)</code> <code class="grammar-paren">(</code><code class="grammar-paren">(</code><a href="#grammar-production-PN_CHARS">PN_CHARS</a> <code class="grammar-alt">|</code> '<code class="grammar-literal">.</code>'<code class="grammar-paren">)</code><code class="grammar-star">*</code> <a href="#grammar-production-PN_CHARS">PN_CHARS</a><code class="grammar-paren">)</code><code class="grammar-opt">?</code></td>
     </tr>
     <tr id="grammar-production-LANG_DIR">
-      <td>[34]</td>
+      <td>[35]</td>
       <td><code>LANG_DIR</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">@</code>' <code class="grammar-brac">[</code><code class="grammar-literal">a-zA-Z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> <code class="grammar-paren">(</code>'<code class="grammar-literal">-</code>' <code class="grammar-brac">[</code><code class="grammar-literal">a-zA-Z0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code><code class="grammar-paren">)</code><code class="grammar-star">*</code> <code class="grammar-paren">(</code>'<code class="grammar-literal">--</code>' <code class="grammar-brac">[</code><code class="grammar-literal">a-zA-Z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code><code class="grammar-paren">)</code><code class="grammar-opt">?</code></td>
     </tr>
     <tr id="grammar-production-INTEGER">
-      <td>[35]</td>
+      <td>[36]</td>
       <td><code>INTEGER</code></td>
       <td>::=</td>
       <td><code class="grammar-brac">[</code><code class="grammar-literal">+-</code><code class="grammar-brac">]</code><code class="grammar-opt">?</code> <code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code></td>
     </tr>
     <tr id="grammar-production-DECIMAL">
-      <td>[36]</td>
+      <td>[37]</td>
       <td><code>DECIMAL</code></td>
       <td>::=</td>
       <td><code class="grammar-brac">[</code><code class="grammar-literal">+-</code><code class="grammar-brac">]</code><code class="grammar-opt">?</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-star">*</code> '<code class="grammar-literal">.</code>' <code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code><code class="grammar-paren">)</code></td>
     </tr>
     <tr id="grammar-production-DOUBLE">
-      <td>[37]</td>
+      <td>[38]</td>
       <td><code>DOUBLE</code></td>
       <td>::=</td>
       <td><code class="grammar-brac">[</code><code class="grammar-literal">+-</code><code class="grammar-brac">]</code><code class="grammar-opt">?</code> <code class="grammar-paren">(</code><code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> '<code class="grammar-literal">.</code>' <code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-star">*</code> <a href="#grammar-production-EXPONENT">EXPONENT</a><code class="grammar-paren">)</code> <code class="grammar-alt">|</code> <code class="grammar-paren">(</code>'<code class="grammar-literal">.</code>' <code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> <a href="#grammar-production-EXPONENT">EXPONENT</a><code class="grammar-paren">)</code> <code class="grammar-alt">|</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> <a href="#grammar-production-EXPONENT">EXPONENT</a><code class="grammar-paren">)</code><code class="grammar-paren">)</code></td>
     </tr>
     <tr id="grammar-production-EXPONENT">
-      <td>[38]</td>
+      <td>[39]</td>
       <td><code>EXPONENT</code></td>
       <td>::=</td>
       <td><code class="grammar-brac">[</code><code class="grammar-literal">eE</code><code class="grammar-brac">]</code> <code class="grammar-brac">[</code><code class="grammar-literal">+-</code><code class="grammar-brac">]</code><code class="grammar-opt">?</code> <code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code></td>
     </tr>
     <tr id="grammar-production-STRING_LITERAL_QUOTE">
-      <td>[39]</td>
+      <td>[40]</td>
       <td><code>STRING_LITERAL_QUOTE</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">&quot;</code>' <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">^</code><code class="grammar-char-escape"><abbr title="dquote">#x22</abbr></code><code class="grammar-char-escape"><abbr title="backslash">#x5C</abbr></code><code class="grammar-char-escape"><abbr title="new line">#x0A</abbr></code><code class="grammar-char-escape"><abbr title="carriage return">#x0D</abbr></code><code class="grammar-brac">]</code> <code class="grammar-alt">|</code> <a href="#grammar-production-ECHAR">ECHAR</a> <code class="grammar-alt">|</code> <a href="#grammar-production-UCHAR">UCHAR</a><code class="grammar-paren">)</code><code class="grammar-star">*</code> '<code class="grammar-literal">&quot;</code>'
@@ -238,7 +244,7 @@
       </td>
     </tr>
     <tr id="grammar-production-STRING_LITERAL_SINGLE_QUOTE">
-      <td>[40]</td>
+      <td>[41]</td>
       <td><code>STRING_LITERAL_SINGLE_QUOTE</code></td>
       <td>::=</td>
       <td>"<code class="grammar-literal">&apos;</code>" <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">^</code><code class="grammar-char-escape"><abbr title="apos">#x27</abbr></code><code class="grammar-char-escape"><abbr title="backslash">#x5C</abbr></code><code class="grammar-char-escape"><abbr title="new line">#x0A</abbr></code><code class="grammar-char-escape"><abbr title="carriage return">#x0D</abbr></code><code class="grammar-brac">]</code> <code class="grammar-alt">|</code> <a href="#grammar-production-ECHAR">ECHAR</a> <code class="grammar-alt">|</code> <a href="#grammar-production-UCHAR">UCHAR</a><code class="grammar-paren">)</code><code class="grammar-star">*</code> "<code class="grammar-literal">&apos;</code>"
@@ -246,37 +252,37 @@
       </td>
     </tr>
     <tr id="grammar-production-STRING_LITERAL_LONG_SINGLE_QUOTE">
-      <td>[41]</td>
+      <td>[42]</td>
       <td><code>STRING_LITERAL_LONG_SINGLE_QUOTE</code></td>
       <td>::=</td>
       <td>"<code class="grammar-literal">&apos;&apos;&apos;</code>" <code class="grammar-paren">(</code><code class="grammar-paren">(</code>"<code class="grammar-literal">&apos;</code>" <code class="grammar-alt">|</code> "<code class="grammar-literal">&apos;&apos;</code>"<code class="grammar-paren">)</code><code class="grammar-opt">?</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">^&apos;\</code><code class="grammar-brac">]</code> <code class="grammar-alt">|</code> <a href="#grammar-production-ECHAR">ECHAR</a> <code class="grammar-alt">|</code> <a href="#grammar-production-UCHAR">UCHAR</a><code class="grammar-paren">)</code><code class="grammar-paren">)</code><code class="grammar-star">*</code> "<code class="grammar-literal">&apos;&apos;&apos;</code>"</td>
     </tr>
     <tr id="grammar-production-STRING_LITERAL_LONG_QUOTE">
-      <td>[42]</td>
+      <td>[43]</td>
       <td><code>STRING_LITERAL_LONG_QUOTE</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">&quot;&quot;&quot;</code>' <code class="grammar-paren">(</code><code class="grammar-paren">(</code>'<code class="grammar-literal">&quot;</code>' <code class="grammar-alt">|</code> '<code class="grammar-literal">&quot;&quot;</code>'<code class="grammar-paren">)</code><code class="grammar-opt">?</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">^&quot;\</code><code class="grammar-brac">]</code> <code class="grammar-alt">|</code> <a href="#grammar-production-ECHAR">ECHAR</a> <code class="grammar-alt">|</code> <a href="#grammar-production-UCHAR">UCHAR</a><code class="grammar-paren">)</code><code class="grammar-paren">)</code><code class="grammar-star">*</code> '<code class="grammar-literal">&quot;&quot;&quot;</code>'</td>
     </tr>
     <tr id="grammar-production-UCHAR">
-      <td>[43]</td>
+      <td>[44]</td>
       <td><code>UCHAR</code></td>
       <td>::=</td>
       <td><code class="grammar-paren">(</code>'<code class="grammar-literal">\u</code>' <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a><code class="grammar-paren">)</code> <code class="grammar-alt">|</code> <code class="grammar-paren">(</code>'<code class="grammar-literal">\U</code>' <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a> <a href="#grammar-production-HEX">HEX</a><code class="grammar-paren">)</code></td>
     </tr>
     <tr id="grammar-production-ECHAR">
-      <td>[44]</td>
+      <td>[45]</td>
       <td><code>ECHAR</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">\</code>' <code class="grammar-brac">[</code><code class="grammar-literal">tbnrf\&quot;&apos;</code><code class="grammar-brac">]</code></td>
     </tr>
     <tr id="grammar-production-NIL">
-      <td>[45]</td>
+      <td>[46]</td>
       <td><code>NIL</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">(</code>' <a href="#grammar-production-WS">WS</a><code class="grammar-star">*</code> '<code class="grammar-literal">)</code>'</td>
     </tr>
     <tr id="grammar-production-WS">
-      <td>[46]</td>
+      <td>[47]</td>
       <td><code>WS</code></td>
       <td>::=</td>
       <td><code class="grammar-char-escape"><abbr title="space">#x20</abbr></code> <code class="grammar-alt">|</code> <code class="grammar-char-escape"><abbr title="horizontal tab">#x09</abbr></code> <code class="grammar-alt">|</code> <code class="grammar-char-escape"><abbr title="carriage return">#x0D</abbr></code> <code class="grammar-alt">|</code> <code class="grammar-char-escape"><abbr title="new line">#x0A</abbr></code>
@@ -284,13 +290,13 @@
       </td>
     </tr>
     <tr id="grammar-production-ANON">
-      <td>[47]</td>
+      <td>[48]</td>
       <td><code>ANON</code></td>
       <td>::=</td>
       <td>'<code class="grammar-literal">[</code>' <a href="#grammar-production-WS">WS</a><code class="grammar-star">*</code> '<code class="grammar-literal">]</code>'</td>
     </tr>
     <tr id="grammar-production-PN_CHARS_BASE">
-      <td>[48]</td>
+      <td>[49]</td>
       <td><code>PN_CHARS_BASE</code></td>
       <td>::=</td>
       <td><code class="grammar-brac">[</code><code class="grammar-literal">A-Z</code><code class="grammar-brac">]</code></td>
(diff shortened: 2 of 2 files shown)


gkellogg:
My interpretation (BNF only) of @afs proposed changes for triple terms and triple occurrences. No change to parser rules, thus far. Raw BNF in Files view, rendered [via GitHack](https://raw.githack.com/w3c/rdf-turtle/triple-term-occurance/spec/turtle-bnf.html).

spec/index.html
```diff
@@ -787,17 +787,38 @@ <h3>Quoted Triples</h3>
       may be the <a data-cite="RDF12-CONCEPTS#dfn-subject">subject</a> or
       <a data-cite="RDF12-CONCEPTS#dfn-object">object</a> of an
       <a data-cite="RDF12-CONCEPTS#dfn-rdf-triple">RDF triple</a>.</p>
+    <p class="ednote"><em>Quoted Triple</em>, if it is retained, may be limited to this spec,
+      or replace with a term such as <em>Triple Token</em>.</p>
```
TallTed, line 791:
```suggestion
      or replaced with a term such as <em>Triple Token</em>.</p>
```

spec/index.html
```diff
@@ -819,19 +840,32 @@ <h3>Quoted Triples</h3>
...
     <section id="annotation-syntax">
       <h2>Annotation Syntax</h2>
       <p>Turtle also defines an <dfn data-lt="annotation-syntax">annotation syntax</dfn>
         to both quote and assert a <a data-cite="RDF12-CONCEPTS#dfn-rdf-triple">triple</a>,
         which provides a convenient shortcut.
-        An annotation can be used to both assert a triple and have that triple be the
+        An annotation can be used to both assert a triple,
+        via an implicit or explicit identifier,
+        and have that triple be the
         <a data-cite="RDF12-CONCEPTS#dfn-subject">subject</a> of further triples.
```
TallTed, line 854: Perhaps a silly question, but can the identifier of an identified triple not also be the _object_ of further triples? If not, why not?

gkellogg: Yes, it can be the subject or object of another triple or triple descriptor.

gkellogg:
The nomenclature and wording in the Quoted Triples section will still require quite a bit of revision. Conceptually, we need to know how to talk about triple descriptors in relation to other triples in a graph, and how quoted triples/triple tokens/triple occurrences related to triple descriptors and what the mean. Most of this needs to go in Concepts, but needs to be echoed in Turtle and other concrete syntaxes. Also, we may discourage the direct use of triple descriptors favoring annotations and quoted triples/whatever.

The main point of this draft, so far, is to get the grammar and basic usage consistent with discussions.

spec/index.html
TallTed:
```suggestion
        <a data-cite="RDF12-CONCEPTS#dfn-subject">subject</a> and/or
        <a data-cite="RDF12-CONCEPTS#dfn-object">object</a> of 
        further triples and/or <a>triple descriptors</a>.
```

gkellogg: Sorry, I neglected what this was associated. The annotation syntax only allows annotating the triple as the subject, not object. Of course, if you use an explicit identifier for the annotation, that can be used as the object of other triples, or a quoted triple consistent with that and using the same identifier could be used as the object of other triples. A quoted triple can be the subject or object of other triples. I believe the identifier of a quoted triple could also be the subject or object of a triple descriptor, but that would be an odd use case. If there is nesting, I would expect only another triple descriptor to be used as the object of a triple descriptor.

TallTed:
```suggestion
        <a data-cite="RDF12-CONCEPTS#dfn-subject">subject</a> of further triples.
        If explicitly identified, the same identifier can then be used as the 
        <a data-cite="RDF12-CONCEPTS#dfn-object">object</a> of additional
        triples and/or <a>triple descriptors</a>.
```

TallTed (changes requested):

spec/index.html
```diff
@@ -846,7 +858,7 @@ <h2>Annotation Syntax</h2>
         to both quote and assert a <a data-cite="RDF12-CONCEPTS#dfn-rdf-triple">triple</a>,
         which provides a convenient shortcut.
         An annotation can be used to both assert a triple,
-        via an implicit or explicit identifier,
+        via an explicit or implicit or explicit identifier,
```
TallTed, line 861: I don't think the order of these matter, but there are only two options, `implicit or explicit` and `explicit or implicit`. `explicit or implicit or explicit` has an extra `explicit`.
```suggestion
        via an explicit or implicit identifier,
```

spec/index.html
```diff
@@ -810,28 +828,56 @@ <h3>Quoted Triples</h3>
...
+        <a href="#grammar-production-predicate"><code>predicate</code></a>, and
+        <a href="#grammar-production-object"><code>object</code></a>
+        to both assert a triple and create a <a>reification</a> composed of any identifier
+        along with the
+        <a href="#grammar-production-subject"><code>subject</code></a>,
+        <a href="#grammar-production-predicate"><code>predicate</code></a>, and
+        <a href="#grammar-production-object"><code>object</code></a>.
       </p>
 
       <p class="note">The annotation syntax is a syntactic short cut in Turtle,
```
pchampin, line 883:
```suggestion
      <p class="note">The annotation syntax is a syntactic shortcut in Turtle,
```

pchampin (dismissed):
I have not finished my review of the Parsing section, but here are already a number of comments.

spec/index.html
```diff
@@ -780,28 +781,45 @@ <h3>Collections</h3>
...
+    <p>A <a>reification</a>
+      is represented using the <a href="#grammar-production-reification"><code>reification</code></a> production
       with <a href="#grammar-production-subject"><code>subject</code></a>,
       <a href="#grammar-production-predicate"><code>predicate</code></a>, and
       <a href="#grammar-production-object"><code>object</code></a>
-      preceded by <a href="#cp-double-lt"><code>&lt;&lt;</code></a>, and
+      preceded by <a href="#cp-double-lt"><code>&lt;&lt;</code></a>
+      along with an optional <a href="#grammar-production-IRIREF"><code>IRIREF</code></a>
+      or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a>
+      delimited by <a href="#cp-pipe"><code>|</code></a>, if present, used as an identifier, and
```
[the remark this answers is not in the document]
TallTed: @pchampin -- I find the above can be read as saying that the identifier and `|` precede the `<<`. I also think that examples need to be closer to *each* description that is then exemplified. Rephrasing a bit, with more markup than most would want...

```suggestion
A reification is represented using the reification production starting with
<a href="#cp-double-lt"><code>&lt;&lt;</code></a>, followed by an optional
identifier (<a href="#grammar-production-IRIREF"><code>IRIREF</code></a> or
<a href="#grammar-production-BlankNode"><code>BlankNode</code></a>) delimited
by <a href="#cp-pipe"><code>|</code></a>, then
<a href="#grammar-production-subject"><code>subject</code></a>,
<a href="#grammar-production-predicate"><code>predicate</code></a>, and
<a href="#grammar-production-object"><code>object</code></a>,
and ending with <a href="#cp-double-gt"><code>&gt;&gt;</code></a>. For
example, `<< :IRIREF | :subject :predicate :object >>`. If the
optional identifier is not present, a fresh RDF blank node is allocated, as
with `<< :subject :predicate :object >>`. </p>
     <p>Note that reifications may be nested, like
`<< :IRIREF1 | :subject1 :predicate1 << :subject2 :predicate2 :object2 >> >>` 
or `<< :subject4 :predicate4 << :IRIREF3 | :subject3 :predicate3 :object3 >> >>`. </p>
```

spec/index.html
```diff
@@ -780,28 +781,45 @@ <h3>Collections</h3>
 
   </section>
 
-  <section id="quoted-triples">
-    <h3>Quoted Triples</h3>
+  <section id="reifications">
+    <h3>Reifications</h3>
```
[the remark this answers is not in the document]
gkellogg: Note added in d32c1b9fd6a178cd27afb651f711d5c00a267689.

spec/index.html
```diff
@@ -794,29 +794,37 @@ <h3>Reifications</h3>
...
       may or may not be asserted within the graph corresponding to this input document.</p>
 
-    <p>A <a>reification</a>
-      is represented using the <a href="#grammar-production-reification"><code>reification</code></a> production
-      with <a href="#grammar-production-subject"><code>subject</code></a>,
+    <p class="note"><a>Reification</a> in RDF 1.2 is a concept distinct from the
+      <a data-cite="RDF12-SEMANTICS#Reif">Reification vocabulary</a> originally
+      defined in <a data-cite="RDF-MT#Reif">RDF Semantics</a>.
+      While both terms describe a representation of an RDF triple using components,
+      RDF 1.2 uses the term to identify a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>
```
TallTed, line 801:
```suggestion
      RDF 1.2 uses the term to identify a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple term</a>
```

TallTed (changes requested):

TallTed (approved):

spec/turtle.bnf
```diff
@@ -23,10 +23,10 @@ String                ::= STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE
...
 PrefixedName          ::= PNAME_LN | PNAME_NS
 BlankNode             ::= BLANK_NODE_LABEL | ANON
-quotedTriple          ::= '<<' qtSubject predicate qtObject '>>'
-qtSubject             ::=	iri | BlankNode | quotedTriple
-qtObject              ::=	iri | BlankNode | literal | quotedTriple
-annotation            ::= '{|' predicateObjectList '|}'
+reifier               ::= '<<' ((iri | BlankNode) '|' )? subject predicate object '>>'
+tripleTerm            ::= '<<(' subject predicate ttObject ')>>'
+ttObject              ::=	iri | BlankNode | literal | tripleTerm
+annotation            ::= '{|' ( (iri | BlankNode) '|' )? predicateObjectList '|}'
```
gkellogg, line 29: Based on recently seen examples that deal with the ambiguity of using `|` to separate the reifier from the predicateObjectList, we could consider using `=>` instead.

```suggestion
annotation            ::= '{|' ( (iri | BlankNode) '=>' )? predicateObjectList '|}'
```

This would also apply to line 26 (where we might want to change the name of the `reifier` production to something like `reifiedTripleTerm`, as `reifier` is used elsewhere to describe `(ire | BlankNode)` in that production.

TallTed:
I think `Allow zero or many annotations` should be `Allow zero or more annotations`, as I don't think there's any forbiddance of *one* annotation, which would seem to be excluded by "many". Possibly `Allow any number of annotations` though that might allow negative numbers, so maybe `Allow any non-negative number of annotations`....

domel:
@gkellogg Could you elaborate on that. Maybe I missed something but why => (two characters) is better than | (one character)?

gkellogg:
> I think `Allow zero or many annotations` should be `Allow zero or more annotations`, as I don't think there's any forbiddance of _one_ annotation, which would seem to be excluded by "many". Possibly `Allow any number of annotations` though that might allow negative numbers, so maybe `Allow any non-negative number of annotations`....

Can you point out where that text is in the document? I don't see it anywhere. That said, other places in this and other documents use "zero or more", so I'd be fine with that.

gkellogg:
> @gkellogg Could you elaborate on that. Maybe I missed something but why => (two characters) is better than | (one character)?

@niklasl pointed out that using `|` creates a problem in SPARQL, where `|` is a property path component, so (in SPARQL) something like `{| a | p o |}` could either be a reifier `a` on the predicate/object `p o`, or a property path `a | p` so changing that separating `|` is something to consider. I've seen `=>` in some more recent notes passed around the group. It's hard to think of another single character that would seem to fit.

Whatever we do for `annotation` will also be necessary for `reifier` to be consistent.

niklasl:
See also https://github.com/w3c/rdf-star-wg/issues/116. One important aspect (IMO) is that a prefix or wrapping notation is valuable for *reading* these, to avoid reading the name (which may be a long hash, as in Wikidata qualifiers) as a predicate (in annotations) or the subject (in the `<< ... >>` form (triple ... descriptors?)).

We also need to be careful with whatever is added so it doesn't block any other future designs that we can foresee, or steps on other syntaxes unnecessarily. In Notation 3 [`=>` is a shorthand](https://w3c.github.io/N3/spec/#n3rules) for `log:implies`.

I've done some more evaluation (all the examples from the UCR plus a gamut of Wikidata data), and actually found that wrapping the name in `|...|` which @afs suggested (among some other alternatives) seems, with proper spacing, to be a fairly OK alternative even in annotations. I'm not sure how many other reasonable wrapping delimiters we have available.

(That's the Ruby/Rust lambda-style; also figuring in some musical and mathematical notations (|abs|), etc. I put [examples in a gist](https://gist.github.com/niklasl/c23925f831950506fde4eb73885319cd).)

I've also suggested some more radical changes; but admittedly some of those caters less for what is *likely* the more common case (one- or many-to-one; names added for reference; annotation data still better to keep with the value). The naming-only form (with any SPARQL-compatible syntax) would still work well if you need to "tag" a bunch of triples with the name of a many-to-many reifier.

domel:
@niklasl totally agree => is usen in N3. IMHO it's a bad choice.

domel:
How about `~`?

gkellogg:
> I've done some more evaluation (all the examples from the UCR plus a gamut of Wikidata data), and actually found that wrapping the name in `|...|` which @afs suggested (among some other alternatives) seems, with proper spacing, to be a fairly OK alternative even in annotations. I'm not sure how many other reasonable wrapping delimiters we have available.

That's pretty much what this version of the grammar does if we allow whitespace in the `{|` and `|}` tokens:

```bnf
annotation            ::= '{' WS* '|' ((iri | BlankNode) '|')? predicateObjectList '|' WS* '}'
```

To make it a token, we'll need to define some terminals:

```bnf
reifid               ::= ((iri | BlankNode) '|')
annotation           ::= ANNO_START reifid? predicateObjectList ANNO_END
ANNO_START           ::= '{' WS* '|'
ANNO_END             ::= '|' WS* '}'
```

I also have a version which defines a `reifid` rule `((iri | BlankNode) '|')`, although I think we'll need to tweak the production names.

niklasl:
> That's pretty much what this version of the grammar does if we allow whitespace in the `{|` and `|}` tokens:
> 
> ```
> annotation            ::= '{' WS* '|' ((iri | BlankNode) '|')? predicateObjectList '|' WS* '}'
> ```

I think that's still ambiguous though, unless using a preceding whitespace is to be significant? Otherwise, this: `{ |:x| :y :z |}` still has the problem of matching as an AlternativePath in SPARQL.

I've tried the suggestion to wrap the name, like:
```
annotation	::=	"{|" embeddedName? predicateObjectList? "|}"
embeddedName	::=	'|' (iri | BlankNode) '|'
```
which parses the examples I linked above. A simplified sample:
```turtle
<Q34851> :nominatedFor <Q103618> {| |<Q103618#6698506f>| a :Nomination ;
        :forWork <Q582281> ;
        :date "1958-02-18"^^xsd:date |} ,
    <Q103618> {| |<Q103618#6698506f>| a :Nomination ;
        :forWork <Q713979> ;
        :date "1959-02-23"^^xsd:date |} .
```
This also helps spotting the identifier in named "quoted" triples  (using more real wikidata to illustrate the problem):
```
<< |s:Q34851-05722875-6765-4486-9197-729D8AB780ED| <Q34851> wd:P4342 "Elizabeth_Taylor_-_filmskuespiller" >>
    wd:P2241 <Q45403344> ;
    :rank :Deprecated .
```
Since otherwise you'd have to scan (as a human reader) beyond the id to know if it's the subject of a triple, or a name followed by a triple. (I've mixed up name and subject in these even when editing my own "toy" examples.)

afs:
> Otherwise, this: `{ |:x| :y :z |}` still has the problem of matching as an AlternativePath in SPARQL.

@niklasl - could you expand on that point please? In SPARQL 1.1 `:s | :p :o ...` is illegal. `|` is contextual - only in paths.

afs:
Regardless of the technical issue, `|` might be considered visually confusing in real queries (with longer prefixes names) and addressed with a two character token. However, we are someway down the "agreed syntax " path - we can't keep chopping and changing.

Let's gather the possibilities and then choose from that list.
 
Please do not allow whitespace inside tokens, especially delimiting tokens `{|` and `|}`, without a very strong need. It is much more restrictive to future enhancements like graph terms.

niklasl:
> > Otherwise, this: `{ |:x| :y :z |}` still has the problem of matching as an AlternativePath in SPARQL.
> 
> @niklasl - could you expand on that point please? In SPARQL 1.1 `:s | :p :o ...` is illegal. `|` is contextual - only in paths.

This is the problem in https://github.com/w3c/rdf-star-wg/issues/116 but with different spacing. So:
```sparql
SELECT * { ?s ?p ?o {| dct:issued | dct:modified "2023" |} . }
```
just becomes:
```sparql
SELECT * { ?s ?p ?o { |dct:issued| dct:modified "2023" |} . }
```
and the problem remains (unless the spacing itself is significant, which would be very hard to spot; and as you note it also conflicts with possible graph literals).

With this:
```sparql
SELECT * { ?s ?p ?o {| |ex:annotation1| dct:issued | dct:modified "2023" |} . }
```
at least it works.

afs:
> This is the problem in [w3c/rdf-star-wg#116](https://github.com/w3c/rdf-star-wg/issues/116) but with different spacing.

An argument for not allowing whitespace in tokens.

The `|...|` bracketing isn't necessary and also it is not clear bracketing.

As soon as the parse is inside the `{|` there are two cases - implicit and explicit reifier. That's a rule of one token lookahead.

For me, the visual case is important and multiple _bracketing_ meanings of `|` isn't ideal.

niklasl:
> An argument for not allowing whitespace in tokens.

Definitely; that was the point.
 
> The `|...|` bracketing isn't necessary and also it is not clear bracketing.
> 
> As soon as the parse is inside the `{|` there are two cases - implicit and explicit reifier. That's a rule of one token lookahead.
> 
> For me, the visual case is important and multiple _bracketing_ meanings of `|` isn't ideal.

I didn't think so either, but since it was among [your suggestions](https://github.com/w3c/rdf-star-wg/issues/116#issuecomment-2072020602) (second bullet, the "naming unit") I thought it worth trying.

> IF we can agree a different reference specifier, which might go in front rather than a separator, THEN good.

Agreed. But since names can be long, "bracketing" them or using a "pseudo-predicate" could be very helpful when reading. (See [above examples](https://github.com/w3c/rdf-turtle/pull/51#issuecomment-2234744655).)

You also advised to use a consistent form of naming both in annotations and the "quoted" triples .(A pseudo-predicate wouldn't work in the latter.)

(Aside: Not sure what to call `<< ... >>` now, as `<<( ...)>>` since the triple terms. You've suggested "triple descriptor" before; which I think works.)

niklasl:
@afs and @domel, you both suggested `~`.

I'm not aware of any use of it in related syntaxes, so I hope it wouldn't co-opt or conflict with anything.

### A. Tilde as a suffix
Annotation form:
```turtle
<Q34851> :nominatedFor <Q103618> {| <Q103618#6698506f> ~ a :Nomination ;
        :forWork <Q582281> ;
        :date "1958-02-18"^^xsd:date |} ,
    <Q103618> {| |<Q103618#6698cb58>| a :Nomination ;
        :forWork <Q713979> ;
        :date "1959-02-23"^^xsd:date |} .
```
Unasserted form:
```turtle
<< <Q103618#6698506f> ~ <Q34851> :nominatedFor <Q103618> >> a :Nomination ;
  :forWork <Q582281> ;
  :date "1958-02-18"^^xsd:date .

<< <Q103618#6698cb58> ~ <Q34851> :nominatedFor <Q103618> >> a :Nomination ;
  :forWork <Q713979> ;
  :date "1959-02-23"^^xsd:date .
```

### B. Tilde as a prefix
Annotation form:
```turtle
<Q34851> :nominatedFor <Q103618> {| ~ <Q103618#6698506f> a :Nomination ;
        :forWork <Q582281> ;
        :date "1958-02-18"^^xsd:date |} ,
    <Q103618> {| |<Q103618#6698cb58>| a :Nomination ;
        :forWork <Q713979> ;
        :date "1959-02-23"^^xsd:date |} .
```
Unasserted form:
```turtle
<< ~ <Q103618#6698506f> <Q34851> :nominatedFor <Q103618> >> a :Nomination ;
  :forWork <Q582281> ;
  :date "1958-02-18"^^xsd:date .

<< ~ <Q103618#6698cb58> <Q34851> :nominatedFor <Q103618> >> a :Nomination ;
  :forWork <Q713979> ;
  :date "1959-02-23"^^xsd:date .
```

### C. Prefix in annotations; suffix in unasserted triples

Not as consistent; then again these are different contexts.

TallTed:
> > [@TallTed] I think [`Allow zero or many annotations`](https://github.com/w3c/rdf-turtle/pull/51/commits/e2715131251bdd817098b5307ccd70ccb8062d4a) should be `Allow zero or more annotations`, as I don't think there's any forbiddance of _one_ annotation, which would seem to be excluded by "many". Possibly `Allow any number of annotations` though that might allow negative numbers, so maybe `Allow any non-negative number of annotations`....
> 
> [@gkellogg] Can you point out where that text is in the document? I don't see it anywhere. That said, other places in this and other documents use "zero or more", so I'd be fine with that.

I've edited my comment, in the quote here and in the original, to have the link. That text doesn't appear in the document, so far as I know; only in the commit comment on https://github.com/w3c/rdf-turtle/pull/51/commits/e2715131251bdd817098b5307ccd70ccb8062d4a (so it poses a potential for confusion only for a few likely readers).

gkellogg:
Another prefix syntax which could be considered is to use `=`. In N3, `=` is a stand-in for `owl:sameAs` and there were some discussions that this could have special meaning as the first part of a annotation (similar to Kurt's proposal for BlankNodePropertyList):

```turtle
<Q34851> :nominatedFor <Q103618> {| = <Q103618#6698506f>; a :Nomination ;
        :forWork <Q582281> ;
        :date "1958-02-18"^^xsd:date |} ,
    <Q103618> {| = <Q103618#6698cb58>; a :Nomination ;
        :forWork <Q713979> ;
        :date "1959-02-23"^^xsd:date |} .
```

Note that this also makes `= <Q103618#6698506f>` it's own predicate-object, except that the same-as nature is folded into the identifier for the annotation (or reified triple term).

spec/turtle.bnf
```diff
@@ -23,10 +23,10 @@ String                ::= STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE
 iri                   ::= IRIREF | PrefixedName
 PrefixedName          ::= PNAME_LN | PNAME_NS
 BlankNode             ::= BLANK_NODE_LABEL | ANON
-quotedTriple          ::= '<<' qtSubject predicate qtObject '>>'
-qtSubject             ::=	iri | BlankNode | quotedTriple
-qtObject              ::=	iri | BlankNode | literal | quotedTriple
-annotation            ::= '{|' predicateObjectList '|}'
+reifier               ::= '<<' ((iri | BlankNode) '|' )? subject predicate object '>>'
```
gkellogg, line 26: In https://github.com/w3c/rdf-concepts/pull/96 the term `reifier` refers to the identifier part of this construct, not the whole construct itself. Alternatives might be:

* `indirectTripleTerm`
* `reifiedTripleTerm`
* `tripleReference`

niklasl: I think `reifiedTripleTerm` is precise?

I thought about `tripleDescriptor` too. Ping @afs &mdash; I think you initially had that as a provisional name for what we now call triple terms. Various definitions exist (["serves to describe or identify"](https://www.merriam-webster.com/dictionary/descriptor), [data descriptor](https://en.wikipedia.org/wiki/Data_descriptor)).  Might be more obscure. Then again, "annotation" isn't clear either, without definition, that it means "a reifier of this asserted triple"; but I still think "annotation" is a good name for the syntax.

*If* `<< ... >>` grows to allow multiple triples inside (["triple blocks"](https://github.com/w3c/rdf-star-wg/issues/116#issuecomment-2234357279)), the name should still work ("simple" vs. "compound"?).

(I agree that these should also work standalone (["reification declaration"](https://github.com/w3c/rdf-star-wg/issues/116#issuecomment-2234357279));  i.e. having the `predicateObjectList` optional.)

niklasl:
> Another prefix syntax which could be considered is to use `=`. In N3, `=` is a stand-in for `owl:sameAs` and there were some discussions that this could have special meaning as the first part of a annotation (similar to Kurt's proposal for BlankNodePropertyList):
> 
> [...]
> 
> Note that this also makes `= <Q103618#6698506f>` it's own predicate-object, except that the same-as nature is folded into the identifier for the annotation (or reified triple term).

Yes; this is what I call the "pseudo-predicate option". *If* chosen, I think it should only be allowed immediately after the opening terminal (i.e. *not* as a special predicate anywhere within).

I'm worried about stepping on N3 by declaring `=` built-in in turtle, but being sugar for `owl:sameAs` in N3. (Unless it's moved away from that? Still, tooling does still use it (e.g. RDFLib).)

[N3 used to have the :- "iso" operator](http://infomesh.net/2002/notation3/) as such. But as [mentioned in the issue](https://github.com/w3c/rdf-star-wg/issues/116#issuecomment-2072150352), it has some problems.

A variant could work, including  `~`. (Or `*` :thinking:  ...)

(I thought `@` worked too, but the old use of it in N3 for prefixing keywords makes me less comfortable with that.)

Does it read well enough for reified triple terms (`<< ... >>`)? Some variants:
```turtle
<< = <Q103618#6698506f> ; <Q34851> :nominatedFor <Q103618> >> a :Nomination .

<< ~ <Q103618#6698506f> ; <Q34851> :nominatedFor <Q103618> >> a :Nomination .

<< * <Q103618#6698506f> ; <Q34851> :nominatedFor <Q103618> >> a :Nomination .
```
(It may be odd to have that semicolon there; but it is also visually distinct, as opposed to just a prefix or suffix operator.)

afs:
I have mocked up the choices discussed below in a SPARQL grammar. I'm not yet sure it is perfect because there are quite a few changes in SPARQL but it is parsing positive test cases. It actually has all the choices - they don't conflict!

SPARQL is more complicated than Turtle because of the way elements fit together e.g. BGPs can end without a `DOT` and be followed by non-triple forms. SPARQL is a superset of Turtle as regards for the subject and also treats predicate-object lists slightly differently.

### Reifier Declarations

There is now a requirement for syntax to being able to declare reifiers without also including triples that use the reifier as subject. Without syntax, the data will have to include mention of `rdf:reifies`.

This removes the need to choose one triple which has `<<>>` as subject and the others jut use the id.

There are two forms:
```
    :s :p :o | :r .
```
_triple_, then optional reifier id, `DOT`.

Putting the reifier id after the triple makes it work in subject predicate-object-list blocks - see below.
It is shorthand for:
```
    :s :p :o .
    :r rdf:reifies <<(:s :p :o )>> .
```
and the quoted triple form:
```
    << :r | :s :p :o >> .
```
being
```
    :r rdf:reifies <<( :s :p :o )>> .
```

This latter form makes `<< :r | :s :p :o >>` behaves in syntax somewhat like `[ :q :x ] .` which is already a special case that can appear on its own.

### Prefix,postfix declaration

This suggests one way to simplify the parser lookahead needs for quoted triples which is to have the reifier id at the end, not the beginning, of quoted triples. The quoted and annotation cases then both are postfix.

```
    <<:s :p :o | :r >>
```

This fits with 
```
    :s :p :o | :r .
```

### Refiier id for Annotation Syntax

Annotation syntax would be 

```
    :s :p :o {| :q :z |} .
```
and with identifier
```
    :s :p :o | :r {| :q :z |} .
```
The `{|..|}` block uses the `|:r` as subject but otherwise is a predicate-object list
giving possibility of use in asserte triple predicate-object list.
```
   :s :p1 :o1 {| :q1 :z1 |}  ;
      :p2 :o2 | :r {| :q2 :z2 |}  ;
      :p3 :o3 | :r ;
      :p4 :o4 .
```
not a two-style form causes by being inside the `{| ... |}`.
```
   :s :p1 :o1 {| :q1 :z1 |}  ;
      :p2 :o2 {| :r | :q2 :z2 |}  ;
      :p3 :o3 | :r ;
      :p4 :o4 .
```
or having `:s :p :o {| :r |} ` for declaration.

### Annotation Syntax Block

Currently `{|...|}`.

We do need an some kind of delimited matched pair for annotation.
- having no end marker makes `;` ambiguous
- using `DOT` (i.e.single, same terminator as other triples sequences) makes end of triples ambiguous
  and annotations can nest (it happens naturally).
- using `|` (i.e a single, different terminator) interacts with nested annotations with reifieir ids.

In SPARQL, only simple paths (single property) can be used for annotation blocks.

SPARQL is more complicated than Turtle because BGPs can end without a `DOT`.

### Choice of reifier id syntax

As current: postfix for annotation, prefix for quoted triple. A lookahead of 2 is needed - which is not impossible (aside: TriG already has a similar situation and there it is complicated).

postfix for quotes triple `<< :s :p :o | :r >>` removes this need for lookahead.

An alternative in an introductory character ([explored above](https://github.com/w3c/rdf-turtle/pull/51#issuecomment-2238898582)). Using `;` as a separator though seems odd to me because `;` is widely used in Turtle already.

```
<< ~ :r | :s :p :o >>
<<  :s :p :o ~ :r >>
:s :p :o ~ :r .
```
Various single characters work, including `|` (even `@` if a space between that and a prefix name is acceptable).

The separating `|` in `<< ~ :r | :s :p :o >>` isn't necessary but avoids a quad-like appearance. Postfix form does have this consideration.

```
    :s :p :o ~ :r  .
    :s :p :o ~ :r {| :q :z |} .
    << :s :p :o  ~ :r >> .
```

`~` is quite light in some fonts. Alternative:

```
<< ~(:r) :s :p :o >>
:s :p :o ~(:r) .
```
The token is `~(` - no white space between the characters.

I'm tending towards `~ :r` at the moment.

gkellogg:
My only concern is that a trailing reifier identifier may be counter-intuitive, but that may just be my own bias from having worked with the `<< :r | :s :p :o >>` syntax for a while. Using `~` in a postfix notation does seem cleaner.

```
:s :p :o ~ :r  .
:s :p :o ~ :r {| :q :z |} .
<< :s :p :o  ~ :r >> .
```

gkellogg:
The annotation syntax get's a bit tricky, if any number of reifiers/annotations can be added to a triple. This allows:

* `:s :p :o ~ :r`
* `:s :p :o ~ :r ~:r1`
* `:s :p :o ~ :r {| :p :q |}`
* `:s :p :o ~ :r {| :p :q |} ~:r1`
* `:s :p :o ~ :r {| :p :q |} ~:r1 {| :p1 :q1 |}`
* `:s :p :o ~ :r {| :p :q |} {| :p1 :q1 |}`
* `:s :p :o {| :p :q |} ~:r1 {| :p1 :q1 |}`

If the grammar has the following rules:

```bnf
objectList            ::= object annotation* ( ',' object annotation* )*
reifier               ::= '~' (iri | BlankNode)
reifiedTriple         ::= '<<' subject predicate object reifier? '>>'
tripleTerm            ::= '<<(' subject predicate ttObject ')>>'
ttObject              ::=	iri | BlankNode | literal | tripleTerm
annotation            ::= reifier? ('{|' predicateObjectList '|}')?
```

There's a conflict because of using both `annotation*` and `reifier? ('{|' predicateObjectList '|}')?` which creates a LL(1) parser conflict. I'm sure there's a cleverer set of rules to avoid this.

niklasl:
I like where this is going. (The SPARQL grammar indeed requires much more care!)

Triple reference `<< :s :p :o ~ :r >> .` with postfix form looks in line with named annotation form (triple comes first).

The `~( name )` form could work with the many-to-one cases: `~( name1, name2 )` (with `,` to avoid confusion with lists, and also not looking like predicate object).

```turtle
:s :p :o {| :q :z |} .

:s :p :o ~(:r) .

:s :p :o ~(:r, _:q) .

<< :s :p :o ~ :r >> .
```

afs:
> My only concern is that a trailing reifier identifier may be counter-intuitive, but that may just be my own bias from having worked with the `<< :r | :s :p :o >>` syntax for a while. Using `~` in a postfix notation does seem cleaner.

Agreed. At one level, it is shame to change what's been written about, but at the same time, it hasn't been universally adopted.
If we go postfix, then `~` vs `|` is pure choice and `~` disconnects from earlier writings and, for me, that is a resoanble decision to make.

afs:
> The annotation syntax get's a bit tricky, if any number of reifiers/annotations can be added to a triple. This allows:
> 
> * `:s :p :o ~ :r`
> * `:s :p :o ~ :r ~:r1`
> * `:s :p :o ~ :r {| :p :q |}`
> * `:s :p :o ~ :r {| :p :q |} ~:r1`
> * `:s :p :o ~ :r {| :p :q |} ~:r1 {| :p1 :q1 |}`
> * `:s :p :o ~ :r {| :p :q |} {| :p1 :q1 |}`
> * `:s :p :o {| :p :q |} ~:r1 {| :p1 :q1 |}`

I think that's good to allow.
That should be possible although not the way that grammar has it.

I can do that for SPARQL once the style is agree and I can trim down the universal grammar. The difference choice start to interact when Writing the multi-occurences does start to mix up with the couple some of the altenrative styles

Tentative direction:
- Declarations for asserted and occurences (whatever the final names are)
- `~:r` reifier ids (or any choice that has a leading marker)
- `~:r` reifier id before an annotation block `{| |}` for one style of explicit reifier ids.

afs:
> Triple reference `<< :s :p :o ~ :r >> .` with postfix form looks in line with named annotation form (triple comes first).

Yes - while its not been the style up to now, the uniformity is appealing.

> The `~( name )` form could work with the many-to-one cases: `~( name1, name2 )` (with `,` to avoid confusion with lists, and also not looking like predicate object).

Firstly, if we have this, we can have  `~ :r` and `~(:r1 r2)`. The comma isn't necessary and, to me, it's odd to have in some places and not others. `,` is object lists (which aren't lists! ... let's not go there). YMMV.

I don't think that `~()` and an annotation block is very helpful especially in SPARQL.
```
    :s :p :o ~(:r1 r2) {| :q :z |} .
    :s :p :o ~(:r1 r2) {| :q ?z |} .
```

These more complex case maybe better as declaration-pattern:
```
    :s :p :o ~ :r .
    :d :e :f ~( :r1 r2 ) .
    :r1 :q :z .
    :r2 :q :z .
```
that is, `~(:r1 r2)` is only allowed in a declaration form.

gkellogg:
> The `~( name )` form could work with the many-to-one cases: `~( name1, name2 )` (with `,` to avoid confusion with lists, and also not looking like predicate object).

I'm not sure that this will be a common enough pattern to create special syntax for it, as it's fairly easy (and arguably clearer) to create separate statements for each reifier. Also, serializing a graph containing duplicate reifiers with some overlapping annotations would be pretty challenging. I'd say we start with the single reifier grammar and re-consider adding `~(name1 name2)` if it becomes important.

niklasl:
> > Triple reference `<< :s :p :o ~ :r >> .` with postfix form looks in line with named annotation form (triple comes first).
> 
> Yes - while its not been the style up to now, the uniformity is appealing.

Agreed.
 
> > The `~( name )` form could work with the many-to-one cases: `~( name1, name2 )` (with `,` to avoid confusion with lists, and also not looking like predicate object).
> 
> Firstly, if we have this, we can have `~ :r` and `~(:r1 r2)`. The comma isn't necessary and, to me, it's odd to have in some places and not others. `,` is object lists (which aren't lists! ... let's not go there). YMMV.

Makes sense.
 
> I don't think that `~()` and an annotation block is very helpful especially in SPARQL.
> 
> ```
>     :s :p :o ~(:r1 r2) {| :q :z |} .
>     :s :p :o ~(:r1 r2) {| :q ?z |} .
> ```

I agree (and find that combination harder to read too).

> These more complex case maybe better as declaration-pattern:
> 
> ```
>     :s :p :o ~ :r .
>     :d :e :f ~( :r1 r2 ) .
>     :r1 :q :z .
>     :r2 :q :z .
> ```
> 
> that is, `~(:r1 r2)` is only allowed in a declaration form.

Yes, I think I'd readily accept that.

It's akin to blank nodes, where the embedded `[ ... ]` form works for simple cases. In complex cases (e.g. many-to-many), declaration patterns probably work better. (Both for serializing and, IMHO, for reading &mdash; I'm used to look for a "top-level description" for names that are used in many places.)

niklasl:
> > The `~( name )` form could work with the many-to-one cases: `~( name1, name2 )` (with `,` to avoid confusion with lists, and also not looking like predicate object).
> 
> I'm not sure that this will be a common enough pattern to create special syntax for it, as it's fairly easy (and arguably clearer) to create separate statements for each reifier. Also, serializing a graph containing duplicate reifiers with some overlapping annotations would be pretty challenging. I'd say we start with the single reifier grammar and re-consider adding `~(name1 name2)` if it becomes important.

Sure; there are pros and cons here (repeating only the object probably isn't too bad).

It might be somewhat important though, so let's keep it open for more feedback. For example, cases derived from Wikidata may map cleanly to multiple reifiers per triple &mdash; here's [a sketch using the `~(name1 ...)` form](https://gist.github.com/niklasl/1845c6bc8b37402cc9698720c2e22f88#file-q34851-rdfstar-namedreifiers-4-ttl-L305) (and two non-asserted triple descriptions at lines 624 and 626).

Some serialization considerations. Given a "random" triple stream (here as pseudo-ntriples-with-pnames):
```turtle
:s :p :o .
r1 rdf:reifies <<( :s :p :o  )>> .
:r2 rdf:type :Note .
:r1 rdf:type :Note .
:r2 rdf:reifies <<( :s :p :o )>> .
:r2 rdf:reifies <<( :s :p :q )>> .
:s :p :q .
```
A process with some memory of seen triples but no buffering nor indexing can still stream out "best effort" Turtle line-wise, making it more "well-formed":
```turtle
:s :p :o ~ :r1 .
:r2 a :Note .
:r1 a :Note .
:s :p :o ~ :r2 .
<< :s :p :q ~ :r2 >> .
:s :p :q .
```
Whereas a pretty-printer with access to the entire graph could do:
```turtle
:s :p :o ~( :r1 :r2 ) ,
    :q ~ :r2 .

:r1 a :Note .

:r2 a :Note .
```
It would, for each triple, serialize all `rdf:reifies` triples as annotation name "markers".

AFAICS, only if such markers are neither reifiers of any other triple, nor the subject of any other triple, can they be serialized using the blank `{| ... |}` annotation syntax. That may very well be in the absolute majority in practice, which is fine; and the syntax caters well for that. I hope these other declaration patterns will also work well for the other, more complex scenarios.

spec/turtle.bnf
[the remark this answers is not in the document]
rat10: Since it is now well established what a `tripleTerm` is, why not take it from there and have `reifiedTerm` to describe what the `reifier` refers to in the baseline proposal. I employed the same logic to name `statedTerm` and `describedTerm` in my recent [proposal](https://lists.w3.org/Archives/Public/public-rdf-star-wg/2024Jul/0109.html) to disambiguate different kinds of tripleTerm instantiations.

tl;dr: `reifiedTerm`

afs:
> The `~( name )` form could work with the many-to-one cases: `~( name1, name2 )`

Am I reading Gregg's multiple annotations proposal correctly here and this can be done with:

> * :s :p :o ~:r ~:r1 .

?

afs:
A question for clarification:
```
:s :p :o ~:r {| :p :q |} {| :p1 :q1 |} .
```
Is this case 1, which I prefer, and was my initial reading 
(... then I wrote the parser rule trying to make it clear what was happening rather than merely passing the language ...)

The second annotation block has a generated reification id and would be the same as writing:
```
:s :p :o {| :p1 :q1 |} ~:r {| :p :q |} .
```
and making
```
:s :p :o {| :p1 :q1 |} {| :p2 :q2 |} .
``` 
two separate blank nodes?
```
:s :p :o .
_:b1 rdf:reifiies <<(:s :p :o)>> .
_:b1 :p1 :q1 .
_:b2 rdf:reifiies <<(:s :p :o)>> .
_:b2 :p2 :q2 .
```

or is it case 2
where the  `~:r` apply to all following blocks until the next reifierId and if so what about
`:s :p :o {| :p1 :q1 |} {| :p2 :q2 |}`. - does the second block "inherit" the blank node from the first?
The "the same bnode" here in case2 feels odd.

At some point, we have to say "don't rely on gnarly expressions to do what you want - write them clearly" and provide a justifiable reading.

Case 1 style would be explaining 
```
:s :p :o {| :p1 :q1 |} {| :p2 :q2 |} .
```
as shorthand for
```
:s :p :o {| :p1 :q1 |} .
:s :p :o {| :p2 :q2 |} .
```

gkellogg:
> ```
> :s :p :o ~:r {| :p :q |} {| :p1 :q1 |} .
> ```
> 
> Is this case 1, which I prefer, and was my initial reading (... then I wrote the parser rule trying to make it clear what was happening rather than merely passing the language ...)
> 
> The second annotation block has a generated reification id and would be the same as writing:
> 
> ```
> :s :p :o {| :p1 :q1 |} ~:r {| :p :q |} .
> ```

That's my interpretation, and what I think makes sense.

> and making
> 
> ```
> :s :p :o {| :p1 :q1 |} {| :p2 :q2 |} .
> ```
> 
> two separate blank nodes?

+1

> ```
> :s :p :o .
> _:b1 rdf:reifiies <<(:s :p :o)>> .
> _:b1 :p1 :q1 .
> _:b2 rdf:reifiies <<(:s :p :o)>> .
> _:b2 :p2 :q2 .
> ```
> 
> or is it case 2 where the `~:r` apply to all following blocks until the next reifierId and if so what about `:s :p :o {| :p1 :q1 |} {| :p2 :q2 |}`. - does the second block "inherit" the blank node from the first? The "the same bnode" here in case2 feels odd.

To me, that doesn't make sense.

> At some point, we have to say "don't rely on gnarly expressions to do what you want - write them clearly" and provide a justifiable reading.
> 
> Case 1 style would be explaining
> 
> ```
> :s :p :o {| :p1 :q1 |} {| :p2 :q2 |} .
> ```
> 
> as shorthand for
> 
> ```
> :s :p :o {| :p1 :q1 |} .
> :s :p :o {| :p2 :q2 |} .
> ```

+1

gkellogg:
There is a bit of ambiguity still in the proposed grammar. `~ :r {| :p :o |}` could also be parsed as `~ :r ~ _:bn {| :p :o |}`, as it's ambiguous if the annotation block stands alone or is intended to use `:r` as it's reifier. One way to solve this would be to define the grammar as follows:

```bnf
annotation            ::= '{|' predicateObjectList '|}'
                        | reifier ('{|' predicateObjectList? '|}')
```

This way `~r` would always need to be followed by a (potentially empty) annotation block. The example above would become `~ :r {||} {| :p :o |}` to generate the following triples:

```turtle
:s :p :o ~r {||} {| :p :o |} .

# expands to

:s :p :o .
:r rdf:reifiies <<(:s :p :o)>> .
_:b1 rdf:reifiies <<(:s :p :o)>> .
_:b1 :p :q .
```

Alternatively, the ambiguity can be resolved in parser logic and use an alternative grammar:

```bnf
annotation            ::= reifier | '{|' predicateObjectList '|}'
```

If a parser parses a reifier and subsequently parses the annotation block it would assign the previously parsed reifier to that annotation block, but the BNF itself is ambiguous which is concerning.

niklasl:
Is there a need to both name and describe the reifier in place? With bnodes its either id or description block, so it would follow the general Turtle design to either id or describe an anonymous reifier here too.

gkellogg:
> Is there a need to both name and describe the reifier in place? With bnodes its either id or description block, so it would follow the general Turtle design to either id or describe an anonymous reifier here too.

We need the ability to name a description block with either an IRI or a blank node. If not provided, the name (reifier) is automatically generated. Because the grammar allows both the description block and the reifier to be optional we have a conflict. Based on discussion, it seems that there _is_ a need to both name and describe or just describe a description block.

afs:
> the BNF itself is ambiguous

The BNF is fine - what has to be defined is the translation from the syntax tree to triples output (section 7).

This is LL(1) for the multiple annotation case via the `*`
```
        Object           :=	GraphNode Annotation
	Annotation       :=	( Reifier | AnnotationBlock )*
	AnnotationBlock  :=	<L_ANN> PropertyList <R_ANN>
```
This does not try to associate the reifier with the annotation block.

It is not showing as ambiguous because the sequence `~:e {| :x :y |}`is just fine as concrete language -  a reifier id followed by a `{| |}` block. 

The meaning, the translation to triples, would have a state variable for the reifier id which is initially unset, then set by `~:e` then cleared by `|}`. Similar to `:s :p1 :o1 ; :p2 :o2 .` passing he suivject on until DOT.

Writing 
```
annotation            ::= '{|' predicateObjectList '|}'
                        | reifier ('{|' predicateObjectList? '|}')?
```
(I think there was a missing `?` on the second line which I've included.)

is a problem for multiple reifiers/annotation blocks. `annotation*` is ambiguous. 

`~:e {| |}` can be first clause, with `?` empty then a second clause, or it can be first clause with non-empty `?`.

afs:
To move forward I suggest moving this PR out of draft so as to merge it to get everything else in to the doc even if the grammar isn't final.

Create a follow-up issue, or issues, for specific points in the grammar.

afs (approved):

domel (approved):

gkellogg:
Squashed and force-pushed to rebase to main.

spec/index.html
```diff
@@ -1627,13 +1701,34 @@ <h3>RDF Term Constructors</h3>
...
+        </td>
+      </tr>
+      <tr id="handle-reifiedTriple"                   ><td style="text-align:left;"            ><a class="type reifiedTriple" href="#grammar-production-reifiedTriple"                 >reifiedTriple                   </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a> | <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a></td>
+        <td>
+          The <a class="type reifiedTriple" href="#grammar-production-reifiedTriple">reifiedTriple</a> production
+          defines an RDF term in addition to a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple term</a>.
+          The term constructed from this production
+          is composed of an identifier from either the <a href="#grammar-production-iri"><code>iri</code></a>
+          or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> productions,
+          if present, otherwise from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
```
TallTed, line 1720:
```suggestion
          The term constructed from this production
          is composed of an optional identifier from either the <a href="#grammar-production-iri"><code>iri</code></a>
          or the <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> productions;
          otherwise, from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
```

spec/index.html
```diff
@@ -1627,13 +1701,34 @@ <h3>RDF Term Constructors</h3>
...
+        </td>
+      </tr>
+      <tr id="handle-annotation"                      ><td style="text-align:left;"            ><a class="type annotation" href="#grammar-production-annotation"                        >annotation                      </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a> | <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a></td>
+        <td>
+          The <a href="#grammar-production-annotation"><code>annotation</code></a> production
+          defines an RDF term in addition to a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple term</a>.
+          The term constructed from this production
+          is composed of an identifier from either the <a href="#grammar-production-iri"><code>iri</code></a>
+          or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> productions,
+          if present, otherwise from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
```
TallTed, line 1730:
```suggestion
          is composed of an optional identifier from either the <a href="#grammar-production-iri"><code>iri</code></a>
          or the <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> productions;
          otherwise, from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
```

TallTed (changes requested):
GitHub suddenly told me that this was merged... an hour ago. *sigh*

I'm thinking it will be faster/easier for you to put these into a new PR than for me, but I can do it if it's a burden.

gkellogg:
> GitHub suddenly told me that this was merged... an hour ago. _sigh_
> 
> I'm thinking it will be faster/easier for you to put these into a new PR than for me, but I can do it if it's a burden.

Sure, I can incorporate this into a followup PR.

spec/index.html
```diff
@@ -780,28 +781,54 @@ <h3>Collections</h3>
...
-      <a href="#grammar-production-object"><code>object</code></a>
-      preceded by <a href="#cp-double-lt"><code>&lt;&lt;</code></a>, and
-      followed by <a href="#cp-double-gt"><code>&gt;&gt;</code></a>.
-      Note that <a data-cite="RDF12-CONCEPTS#dfn-quoted-triple">quoted triples</a>
-      may be nested.
-     </p>
-
-    <pre id="ex-quoted-triple"
+      <a href="#grammar-production-ttObject"><code>ttObject</code></a>,
+      optionally follwed by a <a href="#grammar-production-reifier"><code>reifier</code></a>
```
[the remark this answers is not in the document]
gkellogg: How about "optionally followed by a <a href="#grammar-production-reifier"><code>reifier</code></a>, composed of a <a href="#cp-tilda"><code title="tilda">~</code></a>followed by an <a href="#grammar-production-IRIREF"><code>IRIREF</code></a> or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a>, and ending with `>>`".

spec/index.html
```diff
@@ -780,28 +781,54 @@ <h3>Collections</h3>
...
+      (<a href="#grammar-production-IRIREF"><code>IRIREF</code></a>
+      or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a>),
+      and ending with <a href="#cp-double-gt"><code>&gt;&gt;</code></a>.
+      For example, `&lt;&lt; :subject :predicate :object ~ :IRIREF &gt;&gt;`.
+      If the optional reifier is not present, a fresh RDF blank node is allocated,
+      as with `&lt;&lt; :subject :predicate :object &gt;&gt;`.</p>
```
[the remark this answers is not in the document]
gkellogg: Not allowed by the grammar.

spec/index.html
```diff
@@ -780,28 +781,54 @@ <h3>Collections</h3>
...
+
+    <p class="note"><a>Reified Triples</a> may be nested,
+      like `&lt;&lt;&nbsp;&nbsp;:subject1&nbsp;:predicate1&nbsp;&lt;&lt;&nbsp;:subject2&nbsp;:predicate2&nbsp;:object2&nbsp;&gt;&gt;&nbsp;~:IRIREF1&nbsp;&gt;&gt;`
+      or `&lt;&lt;&nbsp;:subject4&nbsp;:predicate4&nbsp;&lt;&lt;&nbsp;&nbsp;:subject3&nbsp;:predicate3&nbsp;:object3&nbsp;~:IRIREF3&nbsp;&gt;&gt;&nbsp;&gt;&gt;`.</p>
```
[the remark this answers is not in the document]
gkellogg: They should not be doubled, but obviously, the purpose of the `&nbsp;` is to keep the triple elements together.

spec/index.html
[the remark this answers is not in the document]
afs: There is no technical problem with `~` used as Ted suggests.  The usage of wantign to allocate the reifier id once when asserted for common use later has up before.

It has uses to allocate the id at that point so that there is one reifier agreed between later updates.

```
:s :p :o ~ .
```
```
INSERT  { ?e :added ?t } WHERE { :s :p :o ~?e . BIND(NOW() as ?t) }
```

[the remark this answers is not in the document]
gkellogg: This comment is in the context of `reifiedTriple`, so I don't see a case where `<< :s :p :o ~ >>` does any more than `<< :s :p :o >>`, and is certainly not allowed by the grammar (at least the grammar that is included now).

In the annotation use case, `:s :p :o ~` could, indeed, allocate a new blank node, although one which cann't be referenced for creating more triples. It could be equivalent to the following:

```turtle
_:bn rdf:refies <<( :s :p :o )>> .
```

Where the `:_:bn` node is freshly allocated. Still not allowed in the existing grammar, and would require something like the following:

```bnf
annotation            ::= (('~' (iri | Blank Node)?) | '{|' predicateObjectList '|}')*
```

IMO, better to have a single `reifier` rule for both reified triples and annotations.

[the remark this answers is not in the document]
afs: In Turtle, yes, `<<:s :p :o >>` and `<<:s :p :o ~>>` amount to the same thing. 

The `:s :p :o ~` also asserts which is the preparation for later use is more motivating. It's more regularity to include it in `<<>>`.

> better to have a single reifier rule for both reified triples and annotations.

Agreed.

pchampin:
This was discussed during the [rdf-star meeting on 26 September 2024](https://www.w3.org/2024/09/26-rdf-star-minutes.html#t04).

pchampin:
This was discussed during the [rdf-star meeting on 26 September 2024](https://www.w3.org/2024/09/26-rdf-star-minutes.html#t04).

<details><summary><i>View the transcript</i></summary>

<h3>syntax for reifiers</h3>
<p><cite>&lt;doerthe&gt;</cite> I  have to leave, sorry</p>
<p><cite>ora:</cite> I think the main point of contention is whether this is prefix or postfix</p>
<p><cite>gkellogg_:</cite> that, and tilda versus pipe or other characters.</p>
<p><cite>ora:</cite> AndyS, you make a point about ease of parsing</p>
<p><cite>AndyS:</cite> not just that. The pipe is already used in SPARQL, although there are ways around that.<br>
<span>… Enrico made the point that in N-Triples, the reifier comes first (in the subject position).</span><br>
<span>… I think that internal consistency in Turtle is more important.</span></p>
<p><cite>tl:</cite> I made a few proposals, including the use of pipe everywhere, and replacing the curly brackets in the annotation syntax.<br>
<span>… I think we should have looked at the problem that way.</span><br>
<span>… I find Enrico's argument about the position in N-Triples irrelevant.</span></p>
<p><cite>ora:</cite> you are saying this is a usability issue.</p>
<p><cite>tl:</cite> yes, it is the interface, it is important to get this right.</p>
<p><cite>niklasl:</cite> I agree, affordances are important, that's why the pipe is tricky because of its use in SPARQL.</p>
<p><cite>pchampin:</cite> I agree that this is turnning into a broad discussion we can't do in a short amount of time.<br>
<span>… We can consider suffix vs. prefix and separately the tokens used.</span></p>
<p><cite>niklasl:</cite> agreed, long prefix makes things hard to read</p>
<p><cite>&lt;ora&gt;</cite> STRAWPOLL: Postfix?</p>
<p><cite>&lt;ora&gt;</cite> +1</p>
<p><cite>&lt;gkellogg_&gt;</cite> +1</p>
<p><cite>&lt;pchampin&gt;</cite> +1</p>
<p><cite>&lt;tl&gt;</cite> +1</p>
<p><cite>&lt;niklasl&gt;</cite> +1</p>
<p><cite>&lt;pfps&gt;</cite> 0</p>
<p><cite>&lt;Dominik_T&gt;</cite> 0</p>
<p><cite>&lt;gtw&gt;</cite> +1</p>
<p><cite>&lt;TallTed&gt;</cite> +1</p>
<p><cite>&lt;AndyS&gt;</cite> +1</p>
<p><cite>&lt;eBremer&gt;</cite> +1</p>
<p><cite>&lt;ktk&gt;</cite> +1</p>
<p><cite>&lt;ktk&gt;</cite> Tpt: are you around?</p>
<p><cite>&lt;Tpt&gt;</cite> I am back</p>
<p><cite>Ora:</cite> there is still the question of which character we choose</p>
<p><cite>ora:</cite> There are arguments against |</p>
<p><cite>ora:</cite> There will re reifirers without annotations blocks and annotation blocks without reifiers</p>
<p><cite>ora:</cite> if you see an annotation block after a reifier, it is related to this reifier so there is some memory needed</p>
<p><cite>&lt;tl&gt;</cite> my 5cents on syntax: <a href="https://lists.w3.org/Archives/Public/public-rdf-star-wg/2024Sep/0073.html" rel="noopener noreferrer">https://<wbr>lists.w3.org/<wbr>Archives/<wbr>Public/<wbr>public-rdf-star-wg/<wbr>2024Sep/<wbr>0073.html</a></p>
<p><cite>AndyS:</cite> it's easier than doing RDF list</p>
<p><cite>gtw:</cite> Have we a concised summary of the various syntaxes?</p>
<p><cite>&lt;AndyS&gt;</cite> <a href="https://github.com/w3c/rdf-turtle/blob/main/spec/turtle.bnf" rel="noopener noreferrer">https://<wbr>github.com/<wbr>w3c/<wbr>rdf-turtle/<wbr>blob/<wbr>main/<wbr>spec/<wbr>turtle.bnf</a></p>
<p><cite>&lt;pchampin&gt;</cite> &lt;&lt; :s :p :o ~ :r &gt;&gt;.</p>
<p><cite>&lt;tl&gt;</cite> Souri asked for that</p>
<p><cite>&lt;niklasl&gt;</cite> I tried to have a bunch of variants appear "naturally" in <a href="https://niklasl.github.io/rdf-docs/presentations/RDF-reifiers-1/" rel="noopener noreferrer">https://<wbr>niklasl.github.io/<wbr>rdf-docs/<wbr>presentations/<wbr>RDF-reifiers-1/</a> Slide 19 uses that form.</p>
<p><cite>tl:</cite> I would like to point this syntax proposal but I thought we would do syntax later : <a href="https://lists.w3.org/Archives/Public/public-rdf-star-wg/2024Sep/0073.html" rel="noopener noreferrer">https://<wbr>lists.w3.org/<wbr>Archives/<wbr>Public/<wbr>public-rdf-star-wg/<wbr>2024Sep/<wbr>0073.html</a></p>
<p><cite>&lt;pchampin&gt;</cite> :s :p :o ~ :r1 ~ :r2 {| :a :b |}.</p>
<p><cite>gkellogg:</cite> you can insert more than 1 annotation or refifier</p>
<p><cite>&lt;pchampin&gt;</cite> :s :p :o ~ :r1 ~ :r2 {| :a :b |} ~ :r3.</p>
<p><cite>gkellogg:</cite> in any order</p>
<p><cite>pchampin:</cite> if there is no reifier before annotations, the reifier is a blank node</p>
<p><cite>AndyS:</cite> what I find odd is that the annotation block have to have at least one predicate object inside</p>
<p><cite>AndyS:</cite> it makes generating this kind of syntax from a program more complicated</p>
<p><cite>&lt;niklasl&gt;</cite> That empty annotation blocks weren't allowed did trip me up in my introductory slide (8) for annotation sugar.</p>
<p><cite>ora:</cite> Both Turtle and SPARQL use predicateObjectList+</p>
<p><cite>&lt;niklasl&gt;</cite> So +1 from me for allowing it. Makes it easier to save hand-edited, unfinished turtle...</p>
<p><cite>&lt;tl&gt;</cite> from my proposal:   { :s1 :p :o . :s2 :p :o  | :r1 } [| :a :b |] .</p>
<p><cite>&lt;AndyS&gt;</cite> :s :p :o ~ :r1 ~ :r2 {| :a :b |} {| :c :d |}</p>
<p><cite>&lt;niklasl&gt;</cite> &lt;s&gt; :p &lt;o&gt; ~ &lt;r1&gt; {| a :Named |} . &lt;s&gt; :p &lt;o&gt; ~ &lt;r1&gt; ~ {| a :NotNamed |} .</p>
<p><cite>AndyS:</cite> Are you suggesting we have an empty annotation block to "cancel" the preceding reifier?</p>
<p><cite>&lt;niklasl&gt;</cite> See above line. :)</p>
<p><cite>gkellogg:</cite> you can do "~ {|" to get a blank node</p>
<p><cite>&lt;tl&gt;</cite> from my proposal:   &lt;| :s :p :o | :r |&gt; :a :b .</p>
<p><cite>tl:</cite> We should keep {} for group of statements, not annotations</p>
<p><cite>tl:</cite> If we change the abstract reified triple to &lt;&lt;| we use pipes everywhere</p>
<p><cite>tl:</cite> That way the pipe would be everywhere we use RDF-*</p>
<p><cite>gkellogg:</cite> I am afraid it collide with N3 where they use | for object paths</p>
<p><cite>gkellogg:</cite> the triple object can be a path, and I believe it can include "|"</p>
<p><cite>gkellogg:</cite> This would be against a bare pipe</p>
<p><cite>&lt;Dominik_T&gt;</cite> gkellogg can you provide a link or an example where in N3 pipe can be used?</p>
<p><cite>pchampin:</cite> I would like to come back to the previous topic, my personal opinion is that ~ without identifier is a bit strange. I would argue it's not ncessary required we can still write ~ []</p>
<p><cite>gkellogg:</cite> A [] now means bnode property list</p>
<p><cite>gkellogg:</cite> If we allow empty annotation blocks, it's also a way to avoid the empty ~</p>
<p><cite>&lt;gtw&gt;</cite> I believe per the current Turtle draft spec, [] would be valid per the reifier rule: `reifier::='~' (iri | BlankNode)?` (via BlankNode)</p>
<p><cite>AndyS:</cite> I think it's a bit confusing because it would be the  only place where you can have [] but not [ propertyObjectList ]</p>
<p><cite>ora:</cite> If we confuse users it's not going to lead to anything good</p>
<p><cite>ora:</cite> We have this think with multiple reifiers and annotations. Is it really relevant?</p>
<p><cite>ora:</cite> I don't want for people to start to write things and getting it wrong</p>
<p><cite>&lt;niklasl&gt;</cite> Pro/con: &lt;s&gt; :p &lt;o&gt; ~ [ :date "2024" ] . # Pro: Regularity, same syntax for bnodes. Con: may be odd in combination with the naming-and-describing pairing mechanism.</p>
<p><cite>ora:</cite> Syntax discussions are often more difficult semantic discussions</p>
<p><cite>&lt;niklasl&gt;</cite> +1 for syntax being more difficult (also: "there is only syntax")</p>
<p><cite>ora:</cite> It would be nice if we can break this up into a series of decisions</p>
<p><cite>ora:</cite> would be nice if somebody take the trouble to figure out which decisions we have to make, we would have examples of the variants</p>
<p><cite>pchampin:</cite> if we keep "&lt;&lt;" we need to keep it consistent with what people expect from the CG</p>
<p><cite>tl:</cite> &lt;&lt; has been used also for asserted things</p>
<p><cite>tl:</cite> what part of history do we refer to when we talk about user assumptions?</p>
<p><cite>&lt;pchampin&gt;</cite> q.</p>
<p><cite>pchampin:</cite> To be clear I said "if we keep" the &lt;&lt;, getting ride of it alltogether is a way to solve the problem</p>
<p><cite>gkellogg:</cite> It would be nice to make a decision, everything depends on it</p>
<p><cite>ora:</cite> it's unfortunate that the syntax PR has been opened for such a long time with not enough attention</p>
<p><cite>ora:</cite> People often take tiny interest on syntax, way less than it is warenteed</p>
<p><cite>ora:</cite> I am open to suggestions how to do this</p>
<p><cite>AndyS:</cite> we should take this offline</p>
<p><cite>ora:</cite> agree we do this offline, in a way we ended up in a place I did not wanted to end, fighting over these things</p>
<p><cite>ora:</cite> I suggest chairs will pick this up and will go from there</p>
<p><cite>&lt;pfps&gt;</cite> which PR?</p>
<p><cite>&lt;pchampin&gt;</cite> <a href="https://github.com/w3c/rdf-turtle/pull/51" rel="noopener noreferrer">w3c/<wbr>rdf-turtle#51</a></p>
<p><cite>&lt;gb&gt;</cite> <a href="https://github.com/w3c/rdf-turtle/pull/51" rel="noopener noreferrer">MERGED Pull Request 51 Grammar updates for triple terms and occurrences. (by gkellogg) [spec:substantive]</a></p>
<p><cite>pchampin:</cite> In the interest of splitting into multiple decisions, I think we can bundle the brackets for triple term, unasserted triples and annotations</p>

<hr /></details>

rat10:
@afs 
> If we go postfix, then `~` vs `|` is pure choice

I hope this is still valid, and it is good to know.

> and `~` disconnects from earlier writings and, for me, that is a resoanble decision to make.

I beg to differ: we may have been working for years on this, but we're still not in the situation where we have to cater for an installed base. We can still do what we want, and we should strive for a design that is coherent and compelling. Updating our examples or getting confused in discussions by examples from different periods is a minor problem compared to users of the finished spec having to deal with the side effects of some tactical decisions _forever_.

PR #53, opened 2024-02-04
fix missing protocols in e-mail IRIs

exaexa:
This follows a bit of search and a stackoverflow question&answer here: https://stackoverflow.com/questions/35501103/are-email-addresses-without-mailto-allowed-as-iris

Even if the example is valid without the protocol prefixes, I believe the reason why it is indeed valid should be explained or referred to, in order to prevent confusion. If that's the case please let me know.

Thanks!

***

### :boom: Error: 502 Bad Gateway :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 4, 2024, 8:06 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL]([object Object])

```
error code: 502
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/rdf-turtle%2353.)._
</details>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -725,7 +725,7 @@
               foaf:name "Bob" ;
               foaf:knows [
                   foaf:name "Eve" ] ;
-              foaf:mbox <bob@example.com> ] .
+              foaf:mbox <mailto:bob@example.com> ] .
           -->
         </pre>
       </td>
@@ -737,7 +737,7 @@
           _:b <http://xmlns.com/foaf/0.1/name> "Bob" .
           _:b <http://xmlns.com/foaf/0.1/knows> _:c .
           _:c <http://xmlns.com/foaf/0.1/name> "Eve" .
-          _:b <http://xmlns.com/foaf/0.1/mbox> <bob@example.com> .
+          _:b <http://xmlns.com/foaf/0.1/mbox> <mailto:bob@example.com> .
           -->
         </pre>
       </td>


exaexa:
PS I just received an automated mail about "As part of this policy, W3C groups must be able to assess the IPR context of contributions made to their repositories."

Thus I explicitly confirm that this contribution is editorial and non-normative, and thus (as the mail says) there's hopefully nothing to do.

I can't determine if there's anything to be done with respect to contributed code licensing here (the repo license doesn't say much), but just to be sure, I hereby waive all copyright claims for the grand total of 14 letters added in this PR. :]

gkellogg (approved):

gkellogg:
Thanks. Without the scheme it is not a valid IRI.

domel (approved):

TallTed (approved):

Tpt (approved):

exaexa:
O nice, thanks all!

PR #62, opened 2024-08-11
Add processor state, term constructor, and triple constructors. Change reifiedTriple to reifingTriple.

gkellogg:
... for `annotationBlock`, `tripleTerm`, and `reifier`.

Also, extract the `annotationBlock` production from the `annotation` production.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/62.html" title="Last updated on Aug 27, 2024, 10:54 PM UTC (15297c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/62/23809b8...15297c9.html" title="Last updated on Aug 27, 2024, 10:54 PM UTC (15297c9)">Diff</a>


spec/index.html
```diff
@@ -1718,6 +1716,13 @@ <h3>Parser State</h3>
         The |curObject| is bound to the
         <a href="#grammar-production-object"><code>object</code></a> and
         <a href="#grammar-production-ttObject"><code>ttObject</code></a> productions.</li>
+      <li id="curReifier"><a data-cite="RDF12-CONCEPTS#dfn-rdf-term">RDF Term</a> |curReifier| —
+        The |curReifier| is bound to the
+        <a href="#grammar-production-reifier"><code>reifier</code></a> and
+        <a href="#grammar-production-annotationBlock"><code>annotationBlock</code></a> productions.</li>
+      <li id="curTripleTerm"><a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple term</a> |curTripleTerm| —
+        Is set in the <a href="#annotation">Annotations</a> constructor.
```
TallTed, line 1724: Maintaining the earlier description pattern
```suggestion
        The |curTripleTerm| is set in the <a href="#annotation">Annotations</a> constructor.
```

spec/index.html
```diff
@@ -1753,6 +1758,16 @@ <h3>RDF Term Constructors</h3>
...
+      <tr id="handle-reifier"                         ><td style="text-align:left;"            ><a class="type reifier" href="#grammar-production-reifier"                              >reifier                         </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-iri">          IRI          </a>
+                                                                                                                                                                                                                                    | <a data-cite="RDF12-CONCEPTS#dfn-blank-node">   blank node   </a></td>
+        <td>
+          The <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a> sets |curReifier| and is taken from the matched
+          <a href="#grammar-production-iri"><code>iri</code></a> production
+          or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> production, if any.
+          If no such production is matched, it is taken
+          from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
```
TallTed, line 1768: "The `term` sets `curReifier` [...]." What's "taken from the matched ... production"? Is that `term` or `curReifier`? I'm guessing here, so this may be wrong.
```suggestion
          The <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a> sets |curReifier| which is taken from the matched
          <a href="#grammar-production-iri"><code>iri</code></a> production
          or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> production, if any.
          If no such production is matched, |curReifier| is taken
          from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
```

TallTed: This might be more correct?
```suggestion
          The |curReifier| is set from <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a>, which is taken from the matched
          <a href="#grammar-production-iri"><code>iri</code></a> production
          or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> production, if any.
          If no such production is matched, <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a> is taken
          from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
```

spec/index.html
```diff
@@ -1762,28 +1777,19 @@ <h3>RDF Term Constructors</h3>
...
+      <tr id="handle-reifiedTriple"                   ><td style="text-align:left;"            ><a class="type reifiedTriple" href="#grammar-production-reifiedTriple"                  >reifiedTriple                   </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a> | <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a></td>
         <td>
-          The <a class="type reifiedTriple" href="#grammar-production-reifiedTriple">reifiedTriple</a> production
-          defines an RDF term in addition to a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple term</a>.
-          The term constructed from this production
-          is composed of an optional identifier from either the <a href="#grammar-production-iri"><code>iri</code></a>
-          production or the <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> production,
-          if present, otherwise from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
+          The <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a> taken from the matched <a href="#grammar-production-reifier"><code>reifier</code></a>, if any,
```
TallTed, line 1782:
```suggestion
          The <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a> is taken from the matched <a href="#grammar-production-reifier"><code>reifier</code></a>, if any,
```

spec/index.html
```diff
@@ -1762,28 +1777,19 @@ <h3>RDF Term Constructors</h3>
...
         <td>
-          The <a href="#grammar-production-annotation"><code>annotation</code></a> production
-          defines an RDF term in addition to a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple term</a>.
-          The term constructed from this production
-          is composed of an identifier from either the <a href="#grammar-production-iri"><code>iri</code></a>
-          or <a href="#grammar-production-BlankNode"><code>BlankNode</code></a> productions,
-          if present, otherwise from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>.
-          <div class="ednote">Requires state to recall any previously defined <a href="#grammar-production-reifier"><code>reifier</code></a>.</div>
+          The <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a> taken from a previously matched <a>reifier</a>, if any,
```
TallTed, line 1788:
```suggestion
          The <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a> is taken from a previously matched <a>reifier</a>, if any,
```

spec/index.html
```diff
@@ -1807,35 +1813,51 @@ <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Term
...
       Finishing the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
-      and restores the recorded values of |curSubject| and |curPredicate|.</p>
+      restores the recorded values of |curSubject| and |curPredicate|.</p>
 
-    <h4 id="reifier" style="padding-bottom:0; margin-bottom:0;"><span>Reifiers:</span></h4>
+    <h4 id="reifier" style="padding-bottom:0; margin-bottom:0;"><span>Reifier:</span></h4>
+
+    <p>Beginning the <a href="#grammar-production-reifier"><code>reifier</code></a> production,
+      if |curReifier| is set, the production yields the RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|.
+      Finishing the <a href="#grammar-production-reifier"><code>reifier</code></a> production
```
TallTed, line 1822:
```suggestion
      Finishing the <a href="#grammar-production-reifier"><code>reifier</code></a> production,
```

spec/index.html
```diff
@@ -1807,35 +1813,51 @@ <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Term
...
       <a href="#grammar-production-predicate"><code>predicate</code></a>, and
       <a href="#grammar-production-object"><code>object</code></a> productions.
-      Finishing the <a href="#grammar-production-reifier"><code>reifier</code></a> production
+      Finishing the <a href="#grammar-production-reifiedTriple"><code>reifiedTriple</code></a> production
       yields the RDF triple |curSubject| <code>rdf:reifies</code> |TT|
       and restores the recorded values of |curSubject| and |curPredicate|.</p>
 
+    <h4 id="annotationBlock" style="padding-bottom:0; margin-bottom:0;"><span>Annotation Block:</span></h4>
+
+    <p>Beginning the <a href="#grammar-production-annotationBlock"><code>annotationBlock</code></a> production
```
TallTed, line 1843:
```suggestion
    <p>Beginning the <a href="#grammar-production-annotationBlock"><code>annotationBlock</code></a> production,
```

spec/index.html
```diff
@@ -1807,35 +1813,51 @@ <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Term
...
+      |curReifier| is taken from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a> if not already set,
```
TallTed, line 1844:
```suggestion
      |curReifier| is taken from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>, if not already set,
```

TallTed (changes requested):
Mostly small stuff. One big question.

TallTed (commented):
Source lines 1764-1768 may still need work — I'm not sure what "set from" means, compared to "set to" and "taken from". This part of the doc may become clearer with walk-through examples, with real values — and may need that help for readers other than me, too.

gkellogg:
> Source lines 1764-1768 may still need work — I'm not sure what "set from" means, compared to "set to" and "taken from". This part of the doc may become clearer with walk-through examples, with real values — and may need that help for readers other than me, too.

I'd like to get some input from @afs before making more changes, as I believe the original style for processor events may be his invention. We have not previously given examples of how to parse documents, and I don't think it makes sense to do so now. My guess is that most parsers tend to do their own thing and not strictly follow these rules, but they're needed to express the intention for how to turn Turtle into the Abstract Syntax.

spec/index.html
```diff
@@ -1753,6 +1758,16 @@ <h3>RDF Term Constructors</h3>
...
       <tr id="handle-collection2"                     >                                                                                                                                                                           <td><a data-cite="RDF12-CONCEPTS#dfn-iri">          IRI          </a></td><td>For empty lists, the resulting IRI is <code>rdf:nil</code>. Note the rules for <code>collection</code> in the next section.</td></tr>
+      <tr id="handle-reifier"                         ><td style="text-align:left;"            ><a class="type reifier" href="#grammar-production-reifier"                              >reifier                         </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-iri">          IRI          </a>
+                                                                                                                                                                                                                                    | <a data-cite="RDF12-CONCEPTS#dfn-blank-node">   blank node   </a></td>
+        <td>
+          The |curReifier| is set from <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a>, which is taken from the matched
```
gkellogg, line 1764:
```suggestion
          The |curReifier| is taken from <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">term</a>, which is taken from the matched
```

TallTed:
Rather than 17 (or more) suggestions, one comment:

`reifing` should be `reifying`, in all cases.

gkellogg:
> Rather than 17 (or more) suggestions, one comment:
> 
> `reifing` should be `reifying`, in all cases, matching the case of the existing occurence.

Fixed! Brain fart in search/replace.

spec/index.html
```diff
@@ -812,15 +812,15 @@ <h3>Triple Terms</h3>
...
 
-    <p>A <dfn data-cite="RDF12-CONCEPTS#dfn-reified-triple">reified triple</dfn>
+    <p>A <dfn data-cite="RDF12-CONCEPTS#dfn-reifying-triple">reifying triple</dfn>
       may be the <a data-cite="RDF12-CONCEPTS#dfn-subject">subject</a> or
       <a data-cite="RDF12-CONCEPTS#dfn-object">object</a> of an
       <a data-cite="RDF12-CONCEPTS#dfn-rdf-triple">RDF triple</a>.</p>
 
-    <p>A <a>reified triple</a> is syntactic sugar for a specific relationship between an
+    <p>A <a>reifying triple</a> is syntactic sugar for a specific relationship between an
       identifier (<dfn data-cite="RDF12-CONCEPTS#dfn-reifier">reifier</dfn>) and a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple term</a>.
```
hartig, line 824: I would consider this sentence as wrong now. As per the definition in PR https://github.com/w3c/rdf-concepts/pull/98, a _reifying triple_ is an RDF triple whose predicate is `rdf:reifies` and whose object is a triple term. I don't think that such an RDF triple can be said to be syntactic sugar.

spec/index.html
```diff
@@ -812,15 +812,15 @@ <h3>Triple Terms</h3>
...
-  <section id="reified-triples">
-    <h3>Reified Triples</h3>
+  <section id="reifying-triples">
+    <h3>Reifying Triples</h3>
 
-    <p>A <dfn data-cite="RDF12-CONCEPTS#dfn-reified-triple">reified triple</dfn>
+    <p>A <dfn data-cite="RDF12-CONCEPTS#dfn-reifying-triple">reifying triple</dfn>
       may be the <a data-cite="RDF12-CONCEPTS#dfn-subject">subject</a> or
       <a data-cite="RDF12-CONCEPTS#dfn-object">object</a> of an
       <a data-cite="RDF12-CONCEPTS#dfn-rdf-triple">RDF triple</a>.</p>
```
hartig, line 821: I would consider this sentence as wrong now.

As per the definition in PR https://github.com/w3c/rdf-concepts/pull/98, a _reifying triple_ is an RDF triple whose predicate is `rdf:reifies` and whose object is a triple term. I don't think the intention here is to say that such an RDF triple may be the subject or object of an(other) RDF triple, but that's exactly what the sentence says now.

hartig:
I am skeptical about the change of `reifiedTriple` to `reifyingTriple`.

I understand that expressions accepted by this production are meant to be parsed into a _reifying triple_ (using the term "reifying triple" here as per the current definition in PR https://github.com/w3c/rdf-concepts/pull/98). But conceptually, at least for me, such an expression is something else than a reifying triple, and calling such an expression a "reifying triple" may be a source for confusion.

For instance, the expression
```
<< ex:bob  foaf:knows  ex:alice >>
```
matches the production. If we were to refer to this expression as a "reifying triple", I guess many users would assume that the triple (`ex:bob`, `foaf:knows`, `ex:alice`) is the one that we refer to as being "reifying", but it is not. Instead, the reifying triple is the one that this expression expands to; namely:
```
_:b rdf:reifies <<( ex:bob  foaf:knows  ex:alice )>> .
```

As a separate but related comment, I think that replacing the term "reified triple" by "reifying triple" as currently done in this PR -- namely, in the first two sentences of [Section 2.10](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/62.html#reifying-triples) -- is wrong. I created two separate review comments for these two sentences ([comment 1](https://github.com/w3c/rdf-turtle/pull/62/files#r1718774786) and [comment 2](https://github.com/w3c/rdf-turtle/pull/62/files#r1718775115)).

spec/index.html
TallTed: Just as `a` can be said to be syntactic sugar for `rdf:type`,
```ttl
PREFIX  ex:  <http://example.com/#>
ex:reifier  rdf:reifies    <<( ex:subject  ex:predicate  ex:object )>>
```
— can be said to be syntactic sugar for —
```turtle
PREFIX  ex:  <http://example.com/#>
ex:reifier  rdf:type       ex:triple_term ;
            rdf:subject    ex:subject ;
            rdf:predicate  ex:predicate ;
            rdf:object     ex:object .
```
Note that this does *not* include the triple `ex:subject ex:predicate ex:object` which is not itself asserted/stated; it is *only* quoted/annotated.

spec/index.html
gkellogg: The concepts are closely matched, and the `reifyingTriple` production can be used in the subject or object positions of another triple. Note that this is for Turtle triples, which result in triples in the abstract syntax which more closely align with the definition in RDF Concepts. Really `<< :s :p :o ~ :i >>` _is_ a reifying triple, as it results in `:i rdf:reifies <<( :s :p :o )>>` which is the Turtle/N-Triples expression closest to that defined in concepts.

Both `<< :s :p :o >> a rdf:Stated` and `:bob :said << :s :p :o >>` would be legitimate uses of that `reifiedTriple` production.

spec/index.html
gkellogg: Can't really say that one is syntactic sugar for the other, although we could say that the first entails the second, given appropriate semantics.

You're correct that both a not asserted or stated without also having `ex:subject ex:predicate ex:object` in the graph.

gkellogg:
> As a separate but related comment, I think that replacing the term "reified triple" by "reifying triple" as currently done in this PR -- namely, in the first two sentences of [Section 2.10](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/62.html#reifying-triples) -- is wrong. I created two separate review comments for these two sentences ([comment 1](https://github.com/w3c/rdf-turtle/pull/62/files#r1718774786) and [comment 2](https://github.com/w3c/rdf-turtle/pull/62/files#r1718775115)).

I disagree, but I'm open to alternatives. We need the concept **reifying triple**, and the `reifyingTriple` production does result in the spelled out form, so seems appropriate to me.

spec/index.html
afs: I agree with Olaf's point. The subject or object of the other triple is the reifier id itself. It is not the `<< :s :p :o >>` item.

In that sense, the triple has been reified and then the reifier is used in the subject or object position.

Therefore I prefer `ReifiedTriple`.

gkellogg: Very well, I change the term back to `reifiedTriple` but refer to "reifying triple" in RDF Concepts.

My point on subject/object was that, in the Turtle syntax, reifiedTriple can be used in either the subject or object position. Of course, this results in triples where it is the reifier that is in the subject or object position.

TallTed: > Really `<< :s :p :o ~ :i >>` _is_ a reifying triple

Rather, it is syntactic sugar for a reifying triple, that being `:i rdf:reifies <<( :s :p :o )>>`.

> Both `<< :s :p :o >> a rdf:Stated` and `:bob :said << :s :p :o >>` would be legitimate uses of that `reifiedTriple` production.

I think that those are uses of sugar for `reifyingTriple` productions, here for `[] rdf:reifies <<( :s :p :o )>> ; a rdf:Stated` and `:bob :said [] . [] rdf:reifies <<( :s :p :o )>>`. (The latter might be incorrect notation for the two instances of what must be the same blankNode, which had no identifier in the sugar statement, and so it seems should have no identifier in the de-sugared statement.)

TallTed:
I am taking this thread out of the context of any single suggestion, because it will get lost there (as have previous conversations).

We need the concepts of both a Reifying Triple and a Reified Triple. 

These are distinct. 

We have been talking about both for quite some time, and still we have gone back-and-forth with identifying the *same thing* as being an instance of one of these concepts and then the other.

A Reifying Triple is a triple that has a predicate of `rdf:reifies`.

A Reified Triple is the object of a triple that has a predicate of `rdf:reifies`.

`<< ex:bob  foaf:knows  ex:alice >>` might be read has having left out the optional `~ :reifier` (and what seems to me to be an optional but necessarily implied `.`) following `ex:alice`, making this the complete expression:
```
<< ex:bob foaf:knows ex:alice ~ [] . >>
```
That's sugar for the actual reifying triple, which is 
```
[] rdf:reifies <<( ex:bob foaf:knows ex:alice . )>> .
```
That reifies the now-reified triple, 
```
ex:bob foaf:knows ex:alice .
```

I hope that's all clear....

gkellogg:
> A Reified Triple is the object of a triple that has a predicate of `rdf:reifies`.

We call that a [Triple Term](https://www.w3.org/TR/rdf12-concepts/#dfn-triple-term).

> `<< ex:bob foaf:knows ex:alice >>` might be read has having left out the optional `~ :reifier` (and what seems to me to be an optional but necessarily implied `.`) following `ex:alice`, making this the complete expression:
> 
> ```
> << ex:bob foaf:knows ex:alice ~ [] . >>
> ```

The following are equivalent:

* `<< ex:bob foaf:knows ex:lice >>`
* `<< ex:bob foaf:knows ex:lice ~ >>`
* `<< ex:bob foaf:knows ex:lice ~ [] >>`

They are all the same as `[] rdf:reifies <<( ex:bob foaf:knows ex:lice )>> .`

Note that the `reifiedTriple` (or whatever we finally settle on) production does not include the `.` in the grammar.

domel (approved):
Now looks good.

spec/index.html
```diff
@@ -812,16 +812,21 @@ <h3>Triple Terms</h3>
...
-      identifier (<dfn data-cite="RDF12-CONCEPTS#dfn-reifier">reifier</dfn>) and a <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple term</a>.
+  <section id="reifying-triples">
+    <h3>Reifying Triples</h3>
+
+    <p>In RDF, <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a> are rarely used directly,
+      as they are generally restricted to be used
+      only as the <a data-cite="RDF12-CONCEPTS#dfn-object">object</a> of a <a data-cite="RDF12-CONCEPTS#dfn-rdf-triple">triple</a>
+      using the `rdf:reifies` <a data-cite="RDF12-CONCEPTS#dfn-predicate">predicate</a>.
+      Such a triple is called a <dfn data-cite="RDF12-CONCEPTS#dfn-reifying-triple">reifying triple</dfn>.
+      Turtle supports <a>reifying triples</a> using the <a href="#grammar-production-reifiedTriple"><code>reifiedTriple</code></a> production.</p>
```
hartig, line 823:
```suggestion
      Turtle provides a shorthand notation for writing <a>reifying triples</a>
      using the <a href="#grammar-production-reifiedTriple"><code>reifiedTriple</code></a> production.</p>
```

spec/index.html
```diff
@@ -812,16 +812,21 @@ <h3>Triple Terms</h3>
...
+
+    <p>A <a href="#grammar-production-reifiedTriple"><code>reifiedTriple</code></a>
+      is syntactic sugar represeneting a <a>reifying triple</a>,
```
hartig, line 826:
```suggestion
      is syntactic sugar representing a <a>reifying triple</a>,
```

spec/index.html
```diff
@@ -854,7 +859,7 @@ <h3>Reified Triples</h3>
       as with `&lt;&lt; :subject :predicate :object &gt;&gt;`,
       or `&lt;&lt; :subject :predicate :object ~ &gt;&gt;`.</p>
 
-    <p class="note"><a>Reified Triples</a> may be nested,
+    <p class="note"><a href="#grammar-production-reifiedTriple"><code>reifiedTriples</code></a>> may be nested,
```
hartig, line 862:
```suggestion
    <p class="note"><a href="#grammar-production-reifiedTriple"><code>reifiedTriples</code></a> may be nested,
```

hartig (approved):
Just a few minor things.

TallTed:
>> A Reified Triple is the object of a triple that has a predicate of `rdf:reifies`.
>
> We call that a [Triple Term](https://www.w3.org/TR/rdf12-concepts/#dfn-triple-term).

In other words, "we call the object of a triple that has a predicate of `rdf:reifies` a [Triple Term](https://www.w3.org/TR/rdf12-concepts/#dfn-triple-term)" instead of calling it the rather more intuitive "Reified Triple".

At the risk of howls of "bike-shedding", that says to me that `rdf:reifies` and `reifiedTriple` should be revisited, in favor of either a different verb or something that's not a verb at all.

spec/turtle.bnf
```diff
@@ -26,11 +26,12 @@ iri                   ::= IRIREF | PrefixedName
 PrefixedName          ::= PNAME_LN | PNAME_NS
 BlankNode             ::= BLANK_NODE_LABEL | ANON
 reifier               ::= '~' (iri | BlankNode)?
-reifiedTriple         ::= '<<' (subject | reifiedTriple) predicate object reifier* '>>'
+reifiedTriple        ::= '<<' (subject | reifiedTriple) predicate object reifier? '>>'
```
afs, line 29:
```suggestion
reifiedTriple        ::= '<<' (subject | reifiedTriple) verb object reifier? '>>'
```

spec/turtle.bnf
```diff
@@ -26,11 +26,12 @@ iri                   ::= IRIREF | PrefixedName
...
 tripleTerm            ::= '<<(' ttSubject predicate ttObject ')>>'
```
afs, line 30:
```suggestion
tripleTerm            ::= '<<(' ttSubject verb ttObject ')>>'
```

spec/index.html
```diff
@@ -1807,35 +1819,51 @@ <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Term
     <p>Beginning the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
       records the |curSubject| and |curPredicate|.
       Finishing the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
-      and restores the recorded values of |curSubject| and |curPredicate|.</p>
+      restores the recorded values of |curSubject| and |curPredicate|.</p>
+
+    <h4 id="reifier" style="padding-bottom:0; margin-bottom:0;"><span>Reifier:</span></h4>
```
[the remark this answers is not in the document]
gkellogg: I agree, I'll remove them.

spec/index.html
```diff
@@ -1807,35 +1819,51 @@ <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Term
...
       is composed from
       the <a href="#grammar-production-subject"><code>subject</code></a>,
       <a href="#grammar-production-predicate"><code>predicate</code></a>, and
       <a href="#grammar-production-object"><code>object</code></a> productions.
-      Finishing the <a href="#grammar-production-reifier"><code>reifier</code></a> production
+      Finishing the <a href="#grammar-production-reifiedTriple"><code>reifiedTriple</code></a> production
       yields the RDF triple |curSubject| <code>rdf:reifies</code> |TT|
       and restores the recorded values of |curSubject| and |curPredicate|.</p>
 
+    <h4 id="annotationBlock" style="padding-bottom:0; margin-bottom:0;"><span>Annotation Block:</span></h4>
```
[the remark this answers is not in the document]
gkellogg: 7.3.5 is needed to handle the `:s :p :o ~ :r` use case, which isn't an AnnotationBlock, but emits a reifying triple. Annotation Block may consume |curReifier| as set in 7.3.2 Reifier. If there is an unconsumed reifier, it is used to emit the reifying triple in Annotations.

I'll switch the two sections, though.

spec/index.html
```diff
@@ -1807,35 +1819,51 @@ <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Term
...
-      A new <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> instance |TT|
-      is created using |curSubject| |curPredicate| |curObject|.
-      The identifier |TI| is taken from the
-      <a href="#handle-annotation"><code>annotation</code></a> term constructor.
+      A new <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> instance |curTripleTerm|
+      is created using |curSubject| |curPredicate| |curObject|,
+      and the value of |curReifier| is reset.
       Finishing the <a href="#grammar-production-annotation"><code>annotation</code></a> production
-      yields the RDF triple |TI| <code>rdf:reifies</code> |TT|
+      yields the RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|,
```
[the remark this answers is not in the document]
gkellogg: As mentioned above, this is needed to emit the reifying triple if |curReifier| was not consumed in an Annotation Block. For example `:s :p :o ~ :r`.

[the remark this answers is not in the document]
afs: OK - I will reread but I thought a reifing triple was created in both parts for `~:r {| :q :z |}`. Maybe Annotation Block only handled the "no explicit reifier case".

And `:s :p :o :~r1 ~:r2 ` .

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
     </p>
 
-    <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Terms:</span></h4>
+    <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Terms</span></h4>
 
     <p>Beginning the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
       records the |curSubject| and |curPredicate|.
       Finishing the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
-      and restores the recorded values of |curSubject| and |curPredicate|.</p>
+      restores the recorded values of |curSubject| and |curPredicate|.</p>
```
afs, line 1822:
```suggestion
      restores the previously recorded values of |curSubject| and |curPredicate|.</p>
```

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
         |curPredicate| <var>N</var> .</span>
     </p>
 
-    <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Terms:</span></h4>
+    <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Terms</span></h4>
 
     <p>Beginning the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
       records the |curSubject| and |curPredicate|.
```
afs, line 1820: Small errata to RDF 1.1:

- It does not say that `curSubject` and `curPredicate` are stacks, or alternative the parser state is a stack.
- It says in 7.3 "The `curSubject` is bound to the [subject](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/62.html#grammar-production-subject) production." True but it is more than just `subject` even in RDF 1.1 (subject position, `blankNodePropertyList`, property lists in the object position and collections). RDF 1.2 adds the `tripleTerm`, `reifiedTriple`, `annotation` and `annotationBlock` cases.

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
       Finishing the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
-      and restores the recorded values of |curSubject| and |curPredicate|.</p>
+      restores the recorded values of |curSubject| and |curPredicate|.</p>
+
+    <h4 id="reifier" style="padding-bottom:0; margin-bottom:0;"><span>Reifier</span></h4>
+
+    <p>Beginning the <a href="#grammar-production-reifier"><code>reifier</code></a> production,
+      if |curReifier| is set, the production yields the RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|.
+      Finishing the <a href="#grammar-production-reifier"><code>reifier</code></a> production,
+      |curReifier| is taken from the <a href="#handle-reifier"><code>reifier</code></a> term constructor.</p>
```
afs, line 1829: Delete this text. Moved to earlier.

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
     <p>Beginning the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
       records the |curSubject| and |curPredicate|.
       Finishing the <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> production
-      and restores the recorded values of |curSubject| and |curPredicate|.</p>
+      restores the recorded values of |curSubject| and |curPredicate|.</p>
+
+    <h4 id="reifier" style="padding-bottom:0; margin-bottom:0;"><span>Reifier</span></h4>
+
+    <p>Beginning the <a href="#grammar-production-reifier"><code>reifier</code></a> production,
+      if |curReifier| is set, the production yields the RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|.
```
afs, line 1827: > if |curReifier| is set,

It is not set until the end of the production so it is that previous sate at this point.

The production will set `curReifier` - there 's no "if". Do this now and not at the end, then yield the reifies triple using `curReifier`.

```suggestion
    |curReifier| is taken from the <a href="#handle-reifier"><code>reifier</code></a> term constructor.
    For the |tripleTerm| from the last RDF triple yielded then yield the the RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|.
```
There is no need to have a stack for `curReifier`. Even in annotations, only the last value is needed.

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
+      Finishing the <a href="#grammar-production-reifier"><code>reifier</code></a> production,
+      |curReifier| is taken from the <a href="#handle-reifier"><code>reifier</code></a> term constructor.</p>
 
-    <h4 id="reifier" style="padding-bottom:0; margin-bottom:0;"><span>Reifiers:</span></h4>
+    <h4 id="reifiedTriple" style="padding-bottom:0; margin-bottom:0;"><span>Reified Triple</span></h4>
 
     <p>Beginning the <a href="#grammar-production-reifiedTriple"><code>reifiedTriple</code></a> production
       records the |curSubject| and |curPredicate|.
```
afs, line 1834: In the parser state section, maybe explain "record" means a stack.

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
       |curSubject| is taken from the
-      <a href="#handle-reifiedTriple"><code>reifiedTriple</code></a> term constructor.
+      <a href="#grammar-production-reifier"><code>reifier</code></a> of the
+      <a href="#handle-reifiedTriple"><code>reifiedTriple</code></a> term constructor
+      if present,
+      or a fresh RDF blank node, if not matched.
       A new <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> instance |TT|
       is composed from
       the <a href="#grammar-production-subject"><code>subject</code></a>,
```
afs, line 1842: It may not be the `subject` production. It can be `reifiedTriple`. They both set `curSubject` but that is used used later as the reifier carried over.

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
       <a href="#grammar-production-predicate"><code>predicate</code></a>, and
       <a href="#grammar-production-object"><code>object</code></a> productions.
-      Finishing the <a href="#grammar-production-reifier"><code>reifier</code></a> production
+      Finishing the <a href="#grammar-production-reifiedTriple"><code>reifiedTriple</code></a> production
       yields the RDF triple |curSubject| <code>rdf:reifies</code> |TT|
```
afs, line 1846: `curSubject` may have been modified by the subject/reifiedTriple in the subject position insiude `<< >>`.
It might be batter use `curReifier` and make it a stack (for annotation and asserted triples it does not need to be a stack as in comment above.

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
 
     <p>Beginning the <a href="#grammar-production-annotation"><code>annotation</code></a> production
       records the |curSubject| and |curPredicate|.
-      A new <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> instance |TT|
-      is created using |curSubject| |curPredicate| |curObject|.
-      The identifier |TI| is taken from the
-      <a href="#handle-annotation"><code>annotation</code></a> term constructor.
+      A new <a href="#grammar-production-tripleTerm"><code>tripleTerm</code></a> instance |curTripleTerm|
+      is created using |curSubject| |curPredicate| |curObject|,
+      and the value of |curReifier| is reset.
```
afs, line 1855:
```suggestion
      and the value of |curReifier| is cleared.           
```
"reset" is to much like "restore".

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
...
-    <p class="ednote">Describe parser state for recording and assigning any <a>reifier</a>.</p>
 
-    <h4 id="propertyList" style="padding-bottom:0; margin-bottom:0;"><span>Property Lists:</span></h4>
+    <h4 id="annotationBlock" style="padding-bottom:0; margin-bottom:0;"><span>Annotation Block</span></h4>
+
+    <p>Beginning the <a href="#grammar-production-annotationBlock"><code>annotationBlock</code></a> production,
+      |curReifier| is taken from a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>, if not already set,
+      and |curSubject| is taken from |curReifier|
+      and yields the RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|,
+      and resets the value of |curReifier|.</p>
```
afs, line 1867:
```suggestion
      if |curReifier| is not set, then |curReifier| is assigned a fresh RDF <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a> and then yield RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|.
      and |curSubject| is taken from |curReifier|
      and yields the RDF triple |curReifier| <code>rdf:reifies</code> |curTripleTerm|,
      and clears the value of |curReifier|.</p>
```
i.e. only yield if curReifier is not set. Previous setting by an explicit reifier already yielded the `rdf:reifies` triple.
This copes with 
```
:s :p :o :~r .
:s :p :o :~r {| ... |} 
:s :p :o {| ... |} :~r .
:s :p :o {| ... |} 
:s :p :o {| 
       :p :q ~:r2 {|  ... |} ;
       :p :q ~:r23{|  ... |} ;
   |} .
:s :p :o ~:r {| ... |}  {| ... |}
```
The last case - the annotates production has a star so clearing `curReifier`on exit does not clear it per loop but it still needs clearing because a reifier may be last.

spec/index.html
gkellogg: Neither `curSubject` nor `curPredicate` (nor the parser state in general) is a stack, they are global state. Triple constructors save and restore their values, so the "stack" is implemented via the activation records of these constructors, such as the Triple Term constructor.

spec/index.html
gkellogg: The text is intended to capture the events generated by `:s :p :o ~:r1 ~:r2`. When seeing `~:r1` there is no previously established `curReifier` so nothing is emitted, then `curReifier` is set from the `reifier` term constructor. The next time when it sees `~:r2`, `curReifier` would be `:r1` so `:ri rdf:reifies <<(...)>>` is emitted and then `curReifier` is set to `:r2`.

There is no stack for `curReifier`, there is just if it was previously set or not (Note that there is no text that says to save the value of `curReifier` anywhere). It is cleared when entering Annotations`, set in either `Reifier` or `Annotation Block`, and consumed in `Reifier`, `Annotation Block` or `Annotations`.

spec/index.html
gkellogg: The reason it is written the way it is is that `Reifier` can set the value that might be used in `Annotation Block`, but we'll only know that if there is an Annotation Block after the Reifier. If it was set, but not used, and we see another Reifier, it can then be emitted for the saved `curReifier`.

spec/index.html
afs: I'm beginning to see what you want to do here but then some text does seem right.

> Finishing the [annotation](https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/62.html#grammar-production-annotation) production yields the RDF triple curReifier rdf:reifies curTripleTerm, resets the value of curReifier

needs a condition that curReifier is set. In `:s :p :o ~:r {| :x :y |}` it is not (it's been used). That's where I see identical triples being emitted once for the  annotationBlock and one at the and of annotations.

There needs to be more recording and restoring:

```
  :s :p :o ~:r1 {| :x1 :y2 ~:r2 {| :xx1 :yy2 |} ;
                   :x2 :y2
                 |}.
```
which I think is:
```
:s :p :o .
:r1 rdf:reifies <<( :s :p :o )>> .
:r1 :x1 :y2 .
:r2 rdf:reifies <<( :r1 :x1 :y2 )>> .
:r2 :xx1 :yy2 .
:r1 :x2 :y2 .
```
note `:r1 :x2 :y2`.

spec/index.html
```diff
@@ -1802,42 +1814,59 @@ <h3>RDF Triples Constructors</h3>
         |curPredicate| <var>N</var> .</span>
     </p>
 
-    <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Terms:</span></h4>
+    <h4 id="tripleTerm" style="padding-bottom:0; margin-bottom:0;"><span>Triple Terms</span></h4>
```
gkellogg, line 1817: I now think that this handler is not needed and can be eliminated.

gkellogg:
@afs, I re-worked the Triple Constructors partly based on your feedback, but it's possible I missed some of your points.

* Removed the "Triple Terms" section, which didn't seem to be needed.
* The Reifier handler now always emits a reifying triple and sets `curReifier`.
* The Annotation Block handler sets `curReifier` if not set, in which case it assigns a blank node and emits the reifying triple.
* The Reified Triples handler sets `curTripleTerm` and produces the node based on `curReifier`.
* The Annotations handler sets `curTripleTerm` and stacks `curReifier` similar to `curSubject` and `curPredicate`.

afs (approved):
The best way forwards is to merge this PR. There are significant updates so making those visible to early implementers will enable feedback.

spec/index.html
gkellogg: That's what my parser produces.

pchampin:
This was discussed during the rdf-star meeting on 2024-08-29:
https://www.w3.org/2024/08/29-rdf-star-minutes.html#t04

Issue #83, opened 2025-02-08
Escape sequences that encode a supplemental code point using a Unicode surrogate pair


peteroupc:
The [RDF 1.1 Turtle test suite](https://github.com/w3c/rdf-tests/tree/main/rdf/rdf11/rdf-turtle) currently contains a test case (`test-38.ttl`) that involves the encoding of a Unicode supplemental code point using a pair of encoded surrogate code points:

    <x> rdfs:label "\ud801\udc69" .

This behavior may be underspecified in the RDF 1.1 Turtle Recommendation and the RDF 1.2 Turtle working draft.  Indeed, in the Recommendation (and the current working draft), "\uxxxx" represents "A Unicode character in the range U+0000 to U+FFFF inclusive", and "\Uxxxxxxxx" represents "A Unicode character in the range U+0000 to U+10FFFF inclusive" and doesn't explicitly forbid the encoding of surrogate points in this manner or representing supplemental code points using pairs of encoded surrogates.

Should this behavior with representing supplemental code points with encoded surrogates be allowed, in both RDF 1.1 and 1.2 Turtle?

By the way, my current implementation rejects "\uxxxx" if a surrogate is represented, whether in pairs or not, since it can't stand alone in a Unicode text string.

gkellogg:
> The [RDF 1.1 Turtle test suite](https://github.com/w3c/rdf-tests/tree/main/rdf/rdf11/rdf-turtle) currently contains a test case (`test-38.ttl`) that involves the encoding of a Unicode supplemental code point using a pair of encoded surrogate code points:
> 
> ```
> <x> rdfs:label "\ud801\udc69" .
> ```

This will need to change, as RDF 1.2 does not allow surrogates. RDF 1.1 had no such restriction.

> This behavior may be underspecified in the RDF 1.1 Turtle Recommendation and the RDF 1.2 Turtle working draft. Indeed, in the Recommendation (and the current working draft), "\uxxxx" represents "A Unicode character in the range U+0000 to U+FFFF inclusive", and "\Uxxxxxxxx" represents "A Unicode character in the range U+0000 to U+10FFFF inclusive" and doesn't explicitly forbid the encoding of surrogate points in this manner or representing supplemental code points using pairs of encoded surrogates.

The restriction comes from RDF Concepts, where an RDF string cannot contain surrogates. This should be noted in Turtle and other syntaxes. 

[...]

PR #84, opened 2025-02-08
Exclude surrogate codepoints

afs:
Say that Uncode code escape sequence must not generate codepoints for surrogates.

I'm not sure this is that best way and it worth getting it right here before including in other documents (NQ, NT, TriG).

The section on the three kinds of escape sequences uses a bulleted list, where the list item has a table.
Now there is more material in these list items, the display does look so good. If this is changed, this too should be sorted out before propagating to other documents.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/84.html" title="Last updated on Feb 20, 2025, 8:30 AM UTC (392e638)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/84/04fb921...392e638.html" title="Last updated on Feb 20, 2025, 8:30 AM UTC (392e638)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -308,7 +308,7 @@
       or <a>prefixed names</a>.
       Relative and resolved IRIs are preceded by <a href="#cp-less-than"><code title="less-than sign">&lt;</code></a>,
       followed by <a href="#cp-greater-than"><code title="greater-than sign">&gt;</code></a>, and
-      may contain <a href="#numeric">numeric escape sequences</a> (described below).
+      may contain <a>numeric escape sequences</a> (described below).
       For example, <code>&lt;http://example.org/#green-goblin&gt;</code>.
     </p>
     <p id="relative-iri"><a data-cite="RDF12-CONCEPTS#dfn-relative-iri">Relative IRI references</a> like <code>&lt;#green-goblin&gt;</code>
@@ -385,7 +385,7 @@
       <ul>
         <li>leading digits, e.g., <code>leg:3032571</code> or <code>isbn13:9780136019701</code></li>
         <li>non leading colons, e.g., <code>og:video:height</code></li>
-        <li><a href="#reserved">reserved character escape sequences</a>, e.g., <code>wgs:lat\-long</code></li>
+        <li><a data-lt="reserved character escape sequence">reserved character escape sequences</a>, e.g., <code>wgs:lat\-long</code></li>
       </ul>
     </div>
 
@@ -466,11 +466,10 @@
         <a href="#cp-apostrophe"><code title="apostrophe">'</code></a>,
         <a href="#cp-triple-quote"><code>"""</code></a>, or
         <a href="#cp-triple-apostrophe"><code>'''</code></a>);
-        a sequence of permitted characters, <a href="#numeric">numeric escapes</a>,
-        and/or <a href="#string">string escapes</a>; and a final delimiter (matching the initial delimiter).
+        a sequence of permitted characters, <a>numeric escape sequences</a>,
+        and/or <a>string escape sequences</a>; and a final delimiter matching the initial delimiter.
         The corresponding <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">RDF lexical form</a>
-        is the characters between the delimiters,
-        after processing any escape sequences.</p>
+        is the characters between the delimiters, after processing any escape sequences.</p>
 
       <p>If present, the <a data-cite="RDF12-CONCEPTS#dfn-language-tag">language tag</a> is preceded by
         an <a href="#cp-at-sign"><code title="at sign">@</code></a>
@@ -637,7 +636,7 @@
     </p>
     <ul>
       <li>The character <a href="#cp-underscore"><code title="underscore">_</code></a> and
-      the digit characters <a href="#cp-zero"><code title="digit zero">0</code></a>–<a href="#cp-nine"><code title="digit nine">9</code></a>, inclusive)
+      the digit characters <a href="#cp-zero"><code title="digit zero">0</code></a>–<a href="#cp-nine"><code title="digit nine">9</code></a>
       may appear anywhere in a
       <a data-cite="RDF12-CONCEPTS#dfn-blank-node-identifier">blank node identifier</a>.</li>
       <li>The character <a href="#cp-full-stop"><code title="full stop">.</code></a> may appear anywhere except the first or last character.</li>
@@ -1174,9 +1173,12 @@
 
   <p>A <a>Turtle document</a> is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a>
     encoded in UTF-8 [[!RFC3629]].
-  Only <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code points</a>
-  in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+10FFFF</code>
-  inclusive are allowed.
+    Only <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>,
+    in the ranges <code class="codepoint">U+0000</code> to <code class="codepoint">U+D7FF</code> 
+    and <code class="codepoint">U+E000</code> to <code class="codepoint">U+10FFFF</code>,
+    are allowed. This excludes 
+    <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">surrogate code points</a>,
+    range <code class="codepoint">U+D800</code> to <code class="codepoint">U+DFFF</code>.
   </p>
 
   <section id="sec-grammar-ws">
@@ -1233,9 +1235,17 @@
 
     <ul>
       <li>
-        <p><em id="numeric">numeric escape sequences</em> represent <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code points</a>:</p>
-
-        <table id="tab-uchar">
+        <p>
+          A <dfn>numeric escape sequence</dfn> represents the value of
+          a <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>.
+        </p>
+        <p>
+          A <a>numeric escape sequence</a> MUST NOT produce a code point value
+          in the range <code class="codepoint">U+D800</code> to <code class="codepoint">U+DFFF</code>,
+          which is the range for 
+          Unicode <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">surrogates</a>.
+        </p>
+        <table id="tab-uchar" class="separated">
           <thead>
             <tr>
               <th>Escape sequence</th>
@@ -1249,11 +1259,11 @@
                 <a href="#grammar-production-HEX"><code>hex</code></a>
                 <a href="#grammar-production-HEX"><code>hex</code></a></td>
               <td>A <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>
-                in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+FFFF</code> inclusive
+                in the ranges <code class="codepoint">U+0000</code> to <code class="codepoint">U+D7FF</code>
+                and <code class="codepoint">U+E000</code> to <code class="codepoint">U+D7FF</code>,
                 corresponding to the value encoded by the four hexadecimal digits interpreted
                 from most significant to least significant digit.</td>
             </tr>
-
             <tr>
               <td><code>\U</code> <a href="#grammar-production-HEX"><code>hex</code></a>
                 <a href="#grammar-production-HEX"><code>hex</code></a>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -1278,6 +1282,11 @@ <h3>Escape Sequences</h3>
             <a href="#grammar-production-HEX"><code>HEX</code></a>
             ::= [0-9] | [A-F] | [a-f]</span></p>
         </blockquote>
+        <p>
+          Unicode <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">surrogate code points</a>,
+          the range <code class="codepoint">U+D800</code> to <code class="codepoint">U+DFFF</code>
+          are excluded.
```
Tpt, line 1288: This prevents paired surrogates to be used (see #83). It sounds like a breaking change from Turtle 1.1 that allowed paired surrogates to be escaped.

spec/index.html
```diff
@@ -1174,9 +1174,13 @@ <h2>Turtle Grammar</h2>
...
-  Only <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code points</a>
-  in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+10FFFF</code>
-  inclusive are allowed.
+    Only <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>,
+    <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code points</a> 
+    in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+10FFFF</code>
+    inclusive, excluding the range <code class="codepoint">U+D800</code> to 
+    <code class="codepoint">U+DFFF</code> 
+    (<a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">surrogate code points</a>),
+    are allowed.
```
Tpt, line 1183: I think this sentence is redundant with the UTF-8 definition that states <q>The definition of UTF-8 prohibits encoding character numbers between U+D800 and U+DFFF</s> but it's maybe a good idea to add it anyway to make things cristal clear.

spec/index.html
afs: The change happened when "RDF string" came in where it uses Unicode scale values.

Given that surrogate pairs are translated by UTF-16 only, if a pair is introduced by escape sequences, I don't see how the 1.1 spec allows for a translation to be done; so they remain raw (and can't be written in UTF-8).

RDF 1.1 Turtle says "Unicode character" - which isn't a defined term (this may be due to improvements in Unicode).
A surrogate isn't what Unicode calls an "abstract character" (i.e. after decoding).

So, for me, this can be treated as errata on RDF 1.1 Turtle.

https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-2/#G14527

>Surrogate
> Permanently reserved for UTF-16; restricted interchange
> Cannot be assigned to abstract character

spec/index.html
afs: Agreed. At one level RDF string is enough.

It would be better written as "the range U+0000 to U+D7FF and U+E000 to U+10FFFF inclusive" because Unicode writes it that way. 

A next sentence saying "no surrogate code points" is then an optional extra for clarity.

spec/index.html
Tpt: Indeed. Thank you for the clarification!

afs: OT: The Wikipedia disambiguation page for "Down the Rabbit Hole" refers ("see also") to the disambiguation page for "Rabbit Hole" ... which in turn refers to the disambiguation page for "Down the Rabbit Hole".  I hope someone did that on purpose.

gkellogg (approved):
We should probably put an entry in the changes section for this.

afs:
> We should probably put an entry in the changes section for this.

Here or in RDF Concepts? Or both?

(added to this PR)

gkellogg:
> > We should probably put an entry in the changes section for this.
> 
> Here or in RDF Concepts? Or both?

if we made grammar or processing changes that affected this, it’s worth noting. Otherwise, if the changes are not normative, it doesn’t need its own entry.

afs:
This second commit 

* Use `<dfn>` for the three kinds escape sequence
* Codepoint range wording changed to "U+0000 to U+D7FF and U+E000 to U+10FFFF" from "whole range except ..."
* Tables have boarders
* There is a change note entry - it can easily be removed if desired

The ids for the `<dfn>` are not preserved. These seem to be internal links for the tables in grammar processing. One was "#string"  and the map use string to bnode used it,when it meant a normal string (bnode labels are not limited to the two chars of a `\t` escape!) The other two were "#reserved" and "#numeric", both of which are not distintive enough.

Not changed: "Encoding considerations:" in the Media Type uses "U+0000 to U+FFFF" etc.

afs:
> if we made grammar or processing changes that affected this, it’s worth noting. Otherwise, if the changes are not normative, it doesn’t need its own entry.

Either is good for me now we see that test-38 was not in the implementation reports.

Whether the escape sequences could be said to be "unclear, on close inspection" or a "erratum" is more a matter of taste.

Slight, mild, preference to include it because this has been quietly rumbling for a long time.

spec/index.html
```diff
@@ -1350,9 +1358,9 @@ <h3>Escape Sequences</h3>
       <thead>
         <tr>
           <th></th>
-          <th><a href="#numeric">numeric<br/>escapes</a></th>
-          <th><a href="#string">string<br/>escapes</a></th>
-          <th><a href="#reserved">reserved character<br/>escapes</a></th>
+          <th><a href="#dfn-numeric-escape-sequence">numeric<br/>escapes</a></th>
```
gkellogg, line 1361: Using data-lt rather than href would be cleaner and make the internal indexing of term usages consistent. `<a data-lt=“numeric escape sequence”>numeric<br/>escapes</a>`.

afs: Done throughout the document.

domel (approved):

spec/index.html
```diff
@@ -466,8 +466,8 @@ <h3>Quoted Literals</h3>
         <a href="#cp-apostrophe"><code title="apostrophe">'</code></a>,
         <a href="#cp-triple-quote"><code>"""</code></a>, or
         <a href="#cp-triple-apostrophe"><code>'''</code></a>);
-        a sequence of permitted characters, <a href="#numeric">numeric escapes</a>,
-        and/or <a href="#string">string escapes</a>; and a final delimiter (matching the initial delimiter).
+        a sequence of permitted characters, a <a>numeric escape sequences</a>,
```
TallTed, line 469:
```suggestion
        a sequence of permitted characters, a <a>numeric escape sequence</a>,
```

spec/index.html
```diff
@@ -1174,9 +1174,11 @@ <h2>Turtle Grammar</h2>
 
   <p>A <a>Turtle document</a> is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a>
     encoded in UTF-8 [[!RFC3629]].
-  Only <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code points</a>
-  in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+10FFFF</code>
-  inclusive are allowed.
+    Only <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>,
+    in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+D7FF</code> 
+    and  <code class="codepoint">U+E000</code> to <code class="codepoint">U+10FFFF</code> inclusive,
+    are allowed.
```
TallTed, line 1180:
```suggestion
    Only <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>
    in the ranges <code class="codepoint">U+0000</code> to <code class="codepoint">U+D7FF</code>
    and <code class="codepoint">U+E000</code> to <code class="codepoint">U+10FFFF</code>, inclusive,
    are allowed.
```

spec/index.html
```diff
@@ -1233,9 +1235,17 @@ <h3>Escape Sequences</h3>
 
     <ul>
       <li>
-        <p><em id="numeric">numeric escape sequences</em> represent <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code points</a>:</p>
-
-        <table id="tab-uchar">
+        <p>
+          A <dfn>numeric escape sequence</dfn> represents the value of
+          a <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>:
```
TallTed, line 1240:
```suggestion
          a <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>.
```

spec/index.html
```diff
@@ -1233,9 +1235,17 @@ <h3>Escape Sequences</h3>
...
+        </p>
+        <p>
+          A <a>numeric escape sequence</a> MUST NOT produce a code point value
+          in the the range <code class="codepoint">U+D800</code> to <code class="codepoint">U+DFFF</code>,
+          which is the range for
```
TallTed, line 1245:
```suggestion
          inclusive, which is the range for 
```

spec/index.html
```diff
@@ -1249,11 +1259,12 @@ <h3>Escape Sequences</h3>
                 <a href="#grammar-production-HEX"><code>hex</code></a>
                 <a href="#grammar-production-HEX"><code>hex</code></a></td>
               <td>A <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>
-                in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+FFFF</code> inclusive
+                in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+D7FF</code> 
+                and <code class="codepoint">U+E000</code> to <code class="codepoint">U+D7FF</code>
+                inclusive
```
TallTed, line 1264:
```suggestion
                in the ranges <code class="codepoint">U+0000</code> to <code class="codepoint">U+D7FF</code>
                and <code class="codepoint">U+E000</code> to <code class="codepoint">U+D7FF</code>,
                inclusive,
```

spec/index.html
```diff
@@ -1264,27 +1275,23 @@ <h3>Escape Sequences</h3>
                 <a href="#grammar-production-HEX"><code>hex</code></a>
                 <a href="#grammar-production-HEX"><code>hex</code></a></td>
               <td>A <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code point</a>
-                in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+0010FFFF</code> inclusive
+                in the range <code class="codepoint">U+0000</code> to 
+                <code class="codepoint">U+D7FF</code> 
+                and <code class="codepoint">U+E000</code> to <code class="codepoint">U+10FFFF</code> inclusive
```
TallTed, line 1280:
```suggestion
                in the ranges <code class="codepoint">U+0000</code> to
                <code class="codepoint">U+D7FF</code>
                and <code class="codepoint">U+E000</code> to <code class="codepoint">U+10FFFF</code>, inclusive,
```

spec/index.html
```diff
@@ -1333,12 +1340,13 @@ <h3>Escape Sequences</h3>
...
 
-      <li>
+      <li> 
         <p>
-          <em id="reserved">reserved character escape sequences</em> consist of
-          a <a href="#cp-backslash"><code title="backslash">\</code></a>
+          A <dfn>reserved character escape sequence</dfn> consists of a
+          <a href="#cp-backslash"><code title="backslash">\</code></a>
           followed by one of <code>~.-!$&amp;'()*+,;=/?#@%_</code>
           and represent the character to the right of the <a href="#cp-backslash"><code title="backslash">\</code></a>.
```
TallTed, line 1351:
```suggestion
          followed by one of these characters <code>~.-!$&amp;'()*+,;=/?#@%_</code>,
          and represents the character to the right of the <a href="#cp-backslash"><code title="backslash">\</code></a>.
```

spec/index.html
```diff
@@ -2224,6 +2232,9 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
     <li>Changes the `LANGTAG` terminal production to
       <a href="#grammar-production-LANG_DIR" class="type langDir"><code>LANG_DIR</code></a> to include
       an optional <a data-cite="RDF12-CONCEPTS#dfn-base-direction">base direction</a>.</li>
+    <li>Clarify that 
+      <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">Unicode surrogates</a> 
+      are not legal in Turtle documents, nor as the value of a Unicode escape sequences.</li>
```
TallTed, line 2237:
```suggestion
      are not legal in Turtle documents, nor as the value of a Unicode escape sequence.</li>
```

TallTed (changes requested):
Fixes of punctuation and singular/plural disagreements

spec/index.html
afs: Fix made, making the all plural.  "sequence of As, Bs, and/or Cs"

spec/index.html
afs: "ranges" fixed.

Note:  Unicode does not say "inclusive" after every range mention so it could be dropped. A range is "a set of numbers that are valid". Intervals are open/closed.

spec/index.html
afs: The table, which is the defintion, makes it clear.

TallTed: All other inclusive ranges in this document include the word "inclusive". I don't see what is saved by not including it here. The fact that this two row table includes the word in both rows makes its lack here even more odd.

afs: There is one other use of "inclusive".

The Media Type registration - which is not part of this PR - does not use it.

(Personally, I'd take them out because Unicode does not use the word.)

TallTed: I think consistency is better than inconsistency. More of these range statements included "inclusive" than did not, so that's the direction I went. Given that it's not used in other specs that discuss Unicode code point ranges, it's probably OK to omit, but I would then do so in all cases. A single statement somewhere (e.g., in a "conventions used in this document" or similar section) saying that all Unicode code point ranges mentioned are inclusive might still be reasonable, to guide and hopefully prevent misunderstandings by future readers.

afs: Changed to say "inclusive" on first use, then not used.

spec/index.html
```diff
@@ -1174,9 +1173,11 @@ <h2>Turtle Grammar</h2>
...
     encoded in UTF-8 [[!RFC3629]].
-  Only <a data-cite="I18N-GLOSSARY#dfn-code-point">Unicode code points</a>
-  in the range <code class="codepoint">U+0000</code> to <code class="codepoint">U+10FFFF</code>
-  inclusive are allowed.
+    Only <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>,
+    in the ranges <code class="codepoint">U+0000</code> to <code class="codepoint">U+D7FF</code> 
+    and <code class="codepoint">U+E000</code> to <code class="codepoint">U+10FFFF</code> inclusive,
+    are allowed.
+    This excludes <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">surrogate code points</a>.
   </p>
```
TallTed, line 1181:
```suggestion
    (This excludes <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">surrogate code points</a>, i.e.,
    the range <code class="codepoint">U+D800</code> to <code class="codepoint">U+DFFF</code>.)
  </p>
```

TallTed (changes requested):
One relatively small change in the middle. A perfect reader will not require this change, but I think it will help most readers, who are not perfect.

spec/index.html
afs: Range for surrogates added.

gkellogg:
I'm merging this after WG discussion. Further issues can be addressed if @TallTed creates a new issue.

PR #88, opened 2025-03-13
Setup dark mode based on work in rdf-primer and elsewhere.

gkellogg:
Some simplifications to table styling were done for simplicity. Not sure that the subtle use of color for tables in light-mode is as effective in dark-mode.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/88.html" title="Last updated on Sep 28, 2025, 3:38 PM UTC (1f34453)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/88/c5be630...1f34453.html" title="Last updated on Sep 28, 2025, 3:38 PM UTC (1f34453)">Diff</a>


domel (approved):

spec/index.html
```diff
@@ -52,14 +53,51 @@
...
+      --mono-3: #5c6370;
+      --hue-1: #56b6c2;
+      --hue-2: #61aeee;
+      --hue-3: #c678dd;
+      --hue-4: #98c379;
+      --hue-5: #e06c75;
+      --hue-5-2: #be5046;
+      --hue-6: #d19a66;
+      --hue-6-2: #e6c07b;
+    }
```
Tpt, line 89: This might lead to bugs in the future if respec/hljs is updated. I don't think we pin respec JS. I am going to open a ticket in ReSpec about it.

spec/index.html
```diff
@@ -83,14 +121,19 @@
...
       .separated thead tr th { border:1px solid black; padding: .2em; min-width: 10vw }
+      .separated tbody tr td:first-child,
+      .separated tbody tr th:first-child {max-width: 220px; overflow-wrap: anywhere;}
       .separated thead tr th:nth-child(2) { word-break: break-all; }
       .separated tbody tr td:nth-child(2) {padding: 3px 2px;}
       .separated tbody tr td:nth-child(3) {padding: 3px 2px;}
       .separated tbody tr td { border:1px solid black; text-align: center; min-width: 10vw }
+
+      body.darkmode .separated thead tr th { border:1px solid white; }
+      body.darkmode .separated tbody tr td { border:1px solid white; }
```
Tpt, line 132: If we are lazy I think we can just write `.separated tbody tr td { border:1px solid; }`. The default border color is black in light mode and white in dark mode (same below)

spec/index.html
niklasl: @tpt That would be great! I failed to act on https://github.com/w3c/rdf-primer/pull/24 and haven't moved this CSS to rdf-common yet. But fixing it in ReSpec would be much better for all! (I probably should have gone that route from the beginning.)

(I might follow with what I've come up with re. highlighting, since ReSpec seems to handle that in a somewhat deprecated manner (CommonJS vs. ESM).)

spec/index.html
gkellogg: Good suggestions, it should be simpler and still look good, now.

spec/index.html
gkellogg: I removed the hljs styling. Until it is applied to ReSpec, example blocks will continue to have a white background in dark mode, but better not tie that style into the HTML on a one-off basis.

niklasl: I commented on https://github.com/speced/respec/issues/4871#issuecomment-2727318995 to mention the tentative solution for the Primer. Unless someone picks it up there, I might have some time for a PR on ReSpec, which as @Tpt says would be for the best (and the only robust solution).

Tpt: I also commented on https://github.com/speced/respec/issues/4871#issuecomment-2727512093 Sadly, fixing properly does not sound trivial.

niklasl: @tpt Important comment! (I agree it's not trivial, alas; but I think you're on to something there!)

niklasl: Aiming to solve this via https://github.com/speced/respec/pull/4925.

spec/index.html
```diff
@@ -52,13 +53,16 @@
   </script>
 
   <style>
+    body { color-scheme: light; }
```
Tpt, line 56: This prevents darkmode to work when JS is not enabled but not sure if it's a big deal.

Tpt (approved):

spec/index.html
```diff
@@ -52,13 +53,16 @@
...
+    body.darkmode { color-scheme: dark; }
```
niklasl, line 57:
```suggestion
```
These should only affect linked SVGs (which this spec does not currently use). Since general support for that will be handled by respec if https://github.com/speced/respec/pull/4925 is merged, removing these lines should happen anyway when that lands, to avoid hard-to-spot regressions (since the upstream fix is not JS-dependent).

spec/index.html
niklasl: Agreed; with the added suggestion to remove those two lines to avoid required further changes (the use of `color-scheme` here should only affect SVGs).

domel:
@niklasl and @Tpt Can this be merged?

TallTed (approved):

PR #90, opened 2025-04-15
Version announcement.

gkellogg:
This goes with https://github.com/w3c/rdf-concepts/pull/187 and uses some of the ideas from [version proposal](https://github.com/w3c/rdf-star-wg/blob/main/docs/version-proposal.md).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/90.html" title="Last updated on Apr 24, 2025, 7:57 PM UTC (7a1756d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/90/c8a7762...7a1756d.html" title="Last updated on Apr 24, 2025, 7:57 PM UTC (7a1756d)">Diff</a>


spec/index.html
```diff
@@ -299,6 +299,66 @@ <h3>Object Lists</h3>
...
+         class="example turtle" data-transform="updateExample"
+         title="Version announcement">
+      <!--
+      @version "1.2" .
+      @prefix : <http://example.com/> .
+      :a :name "Alice" ~ :t {| :statedBy :bob ; :recorded "2021-07-07"^^xsd:date |} .
+      -->
+    </pre>
+
+    <p>Alternatively, the SPARQL style directive can also be used
```
TallTed, line 325:
```suggestion
    <p>Alternatively, the SPARQL style directive can be used
```

spec/index.html
```diff
@@ -299,6 +299,66 @@ <h3>Object Lists</h3>
...
+         title="HTTP version announcement">
+      <!--
+      GET /document.ttl HTTP/1.1
+      Host: example.com
+      Accept: text/turtle; version=1.2
+      -->
+    </pre>
+
+    <p>In the absense of an explicit version announcement,
+      processors should treat the document as having version "1.2".</p>
```
TallTed, line 352: Shouldn't the "unversioned" be treated as 1.0/1.1 (which are handled the same), and only treat the version as 1.2 when declared as such?
```suggestion
    <p>In the absense of an explicit version announcement,
      processors should treat the document as having version "1.1".</p>
```

gkellogg: I don't think this is something we've agreed on. By default, Turtle 1.2 parsers should be able to parse all the 1.2 features. The directive exists so that pre-1.2 parsers can detect early that the document uses features they won't be able to parse.

TallTed: The lack of a version declaration is what is found in all existing data, which is RDF 1.0 or 1.1, and so it should be treated.

The *presence* of a version declaration begins with RDF 1.2, but I think (hope) it should cause no harm if `version 1.1` is added to existing data.

I think pre-1.2 parsers will need to be updated to react appropriately when encountering a version 1.2 declaration, but will need no update to appropriately handle undeclared version 1.1. I don't know what these parsers will do if they encounter a version 1.1 declaration.

spec/index.html
```diff
@@ -299,6 +299,66 @@ <h3>Object Lists</h3>
...
+         class="example turtle" data-transform="updateExample"
+         title="SPARQL-style version announcement">
+      <!--
+      VERSION "1.2"
+      PREFIX : <http://example.com/>
+      :a :name "Alice" ~ :t {| :statedBy :bob ; :recorded "2021-07-07"^^xsd:date |} .
+      -->
+    </pre>
+
+    <p>When providing content over HTTP, serveres can announce the version
```
kasei, line 338:
```suggestion
    <p>When providing content over HTTP, servers can announce the version
```

spec/index.html
```diff
@@ -299,6 +299,66 @@ <h3>Object Lists</h3>
...
+      GET /document.ttl HTTP/1.1
+      Host: example.com
+      Accept: text/turtle; version=1.2
+      -->
+    </pre>
+
+    <p>In the absense of an explicit version announcement,
+      processors should treat the document as having version "1.2".</p>
+
+    <p class="note">Version numbers other thann "1.2" are not defined by
```
kasei, line 354:
```suggestion
    <p class="note">Version numbers other than "1.2" are not defined by
```

spec/index.html
gkellogg: Any existing parsers that follow the 1.1 grammar will fail with a version declaration, other than in Content-Type. It would be silly to update a 1.1 parser part way to 1.2 just to parse `@version` or `VERSION` and not the other constructs. I think we need to discuss this further with the group.

spec/index.html
```diff
@@ -1445,8 +1558,10 @@ <h3>Grammar</h3>
...
         <li>
-          The strings '<a href="#grammar-production-prefixID"><code class="grammar-literal">@prefix</code></a>'
-          and '<a href="#grammar-production-base"><code class="grammar-literal">@base</code></a>'
+          The strings
+          '<a href="#grammar-production-prefixID"><code class="grammar-literal">@prefix</code></a>',
+          '<a href="#grammar-production-base"><code class="grammar-literal">@base</code></a>', and
+          '<a href="#grammar-production-version"><code class="grammar-literal">@version</code></a>'
           match the pattern for <a href="#grammar-production-LANG_DIR"><code>LANG_DIR</code></a>,
           though neither <code class="grammar-literal">prefix</code>
           nor <code class="grammar-literal">base</code>
```
kasei, line 1567:
```suggestion
          though neither <code class="grammar-literal">prefix</code>,
          <code class="grammar-literal">base</code>, nor <code class="grammar-literal">version</code>
```

spec/index.html
```diff
@@ -1786,6 +1905,7 @@ <h3>RDF Term Constructors</h3>
...
+      <tr id="handle-VERSION_SPECIFIER"               ><td style="text-align:left;"            ><a class="type string"       href="#grammar-production-VERSION_SPECIFIER"               >VERSION_SPECIFIER               </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-literal">      literal      </a></td><td>The characters between the <a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>s are taken to form the decimal value of |curVersion|.</td></tr>
```
kasei, line 1908: I think there's a tension here between using a literal form (`"1.2"`) in the surface syntax and a decimal in the parser state definition. If it's really just a decimal, why not a decimal in the surface syntax? On the other hand, I think it might be useful to allow more than just decimals. For example:
* `@version "1.1-rdf-star-cg"` for data created for a system implementing the CG's version of rdf-star which might be syntactically valid RDF 1.2, but doesn't match the semantics
* `@version "1.3-dev` for in-development features that don't yet have agreed upon semantics, but allowing safe, cross-implementation interop experimentation

kasei:
While not changed in this PR, the start of section 7.2 RDF Term Constructors is:

> This table maps productions and lexical tokens to RDF terms or components of RDF terms listed in 7. Parsing :

I don't think this is true anymore with the inclusion of `VERSION_SPECIFIER`. The table suggests that `VERSION_SPECIFIER` produces a literal, but I think that's a generous description insofar as that produced literal would be immediately parsed as a decimal (used to set `curVersion`) and then thrown away.

spec/index.html
kasei: Also, if we want to support a 1.2-subset allowing langdirs but no triple terms… that would probably be more than just a decimal.

gkellogg: Good argument that |curVersiob| should remain a string. But, other values are hypothetical are hypothetical at this point.

spec/index.html
rubensworks: I also remember the sentiment in our last call being that explicitly mentioning version 1.2 should be optional (SHOULD), even if new 1.2 features are used.

spec/index.html
```diff
@@ -299,6 +299,66 @@ <h3>Object Lists</h3>
...
+
+    <p>In the absense of an explicit version announcement,
+      processors should treat the document as having version "1.2".</p>
+
+    <p class="note">Version numbers other than "1.2" are not defined by
+      this specification.
+      Processors may treat this as an error or warning.
+      While a version of "1.1" is not defined, processors may treat this
+      as an indication that no RDF 1.2 features are used, so that
+      if they are encountered when parsing a document, it can be treated as an error.</p>
```
rubensworks, line 359: I'm not sure we need this part. I feel it could introduce more confusion than needed.
Publishers may interpret this as being allowed to specify version 1.1, while this will break older 1.1 parsers.

spec/index.html
```diff
@@ -869,10 +956,21 @@ <h3>Reifying Triples</h3>
...
+      directives early in the document.
+      This allows parsers that do not support these features to detect
+      the presense of such features early.
+    </p>
+
     <pre id="ex-reified-triple"
          class="example turtle" data-transform="updateExample"
          title="Reified Triple">
       <!--
+      VERSION     "1.2"
```
rubensworks, line 973: There was a suggestion in https://github.com/w3c/rdf-star-wg/issues/141 to allow `VERSION 1.2` (without quotes) as shorthand.
Do we want to introduce this in the spec?
If so, it may be useful to use this shorthand in all examples.

spec/index.html
rubensworks: +1 on `xsd:string`

rubensworks (changes requested):

spec/index.html
gkellogg: I'll update the grammar terminal to allow `([0-9]+ '.' [0-9]+) | ('"' [^#x22#x5C#x0A#x0D] '"')` with the provision that the pure-decimal version be treated as a string when setting |curVersion|.

spec/index.html
Tpt: Instead of allowing any string in quotes, what about allowing any sequence of ascii digits and lower case leter separated by `.` or `-`? The grammar would be something like `[0-9a-z]+ ([.-] [0-9a-z]+)*`

This way one can write `@version 1.2 .` or `@version 1.1-rdf-star-cg .` or `@version 1.3-alpha.4 .` 
I find it slightly nicer than strings in quotes while still allowing a big range of values.

gkellogg:
> While not changed in this PR, the start of section 7.2 RDF Term Constructors is:
> 
> > This table maps productions and lexical tokens to RDF terms or components of RDF terms listed in 7. Parsing :
> 
> I don't think this is true anymore with the inclusion of `VERSION_SPECIFIER`. The table suggests that `VERSION_SPECIFIER` produces a literal, but I think that's a generous description insofar as that produced literal would be immediately parsed as a decimal (used to set `curVersion`) and then thrown away.

Updating to describe the production as returning a _literal_ with datatype `xsd:string`, which is used to set |curVersion|.

spec/index.html
```diff
@@ -335,8 +335,22 @@ <h3>Version Announcement</h3>
       -->
     </pre>
 
-    <p>When providing content over HTTP, servers can announce the version
-      using the optional `version` Content-Type parameter:</p>
+    <p>If the value has the form of an xsd:decimal, the surrounding
+      <a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>s
```
TallTed, line 339: '<a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>s' doesn't read well.
'quotation marks (<a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>)' reads better.
```suggestion
      quotation marks (<a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>)
```

spec/index.html
```diff
@@ -335,8 +335,22 @@ <h3>Version Announcement</h3>
...
+         class="example turtle" data-transform="updateExample"
+         title="Quoteless version announcement">
+      <!--
+      VERSION 1.2
+      PREFIX : <http://example.com/>
+      :a :name "Alice" ~ :t {| :statedBy :bob ; :recorded "2021-07-07"^^xsd:date |} .
+      -->
+    </pre>
+
+    <p>When providing content over HTTP, serveres can announce the version
```
TallTed, line 352:
```suggestion
    <p>When providing content over HTTP, servers can announce the version
```

spec/turtle.bnf
```diff
@@ -81,3 +83,5 @@ PLX               ::= PERCENT | PN_LOCAL_ESC
 PERCENT           ::= '%' HEX HEX
 HEX               ::= ([0-9] | [A-F] | [a-f])
 PN_LOCAL_ESC      ::= '\' ( '_' | '~' | '.' | '-' | '!' | '$' | '&' | "'" | '(' | ')' | '*' | '+' | ',' | ';' | '=' | '/' | '?' | '#' | '@' | '%' )
+VERSION_SPECIFIER ::= (    [0-9a-z]+ ([.-] [0-9a-z]+)*    )
+                    | ('"' [0-9a-z]+ ([.-] [0-9a-z]+)* '"')
```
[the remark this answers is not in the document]
kasei: > Open question: If we restrict the syntax inside of the double quote to [0-9a-z]+ ([.-] [0-9a-z]+)*, is there still a point to allow the version identifier to be enclosed in double quote?

This would make it much harder to parse. I personally don't see a lot of value in having a version with no quotes or in an overly restrictive grammar production, and would prefer the grammar just follow `VERSION` (or `@version`) with `String`, with semantic checks on the contents of that string being left to higher-level code.

rubensworks (approved):

spec/turtle.bnf
```diff
@@ -81,3 +83,5 @@ PLX               ::= PERCENT | PN_LOCAL_ESC
 PERCENT           ::= '%' HEX HEX
 HEX               ::= ([0-9] | [A-F] | [a-f])
 PN_LOCAL_ESC      ::= '\' ( '_' | '~' | '.' | '-' | '!' | '$' | '&' | "'" | '(' | ')' | '*' | '+' | ',' | ';' | '=' | '/' | '?' | '#' | '@' | '%' )
+VERSION_SPECIFIER ::= (    [0-9a-z]+ ([.-] [0-9a-z]+)*    )
```
ektrah, line 86: This seems to introduce some ambiguity.

For example, should `1.x:y` be tokenized as `<VERSION_SPECIFIER:'1.x'> <PNAME_LN:':y'>` or as `<INTEGER:'1'> <'.'> <PNAME_LN:'x:y'>`?

If the former (longest match), this looks like a breaking change to me.

gkellogg: I don't think it's quite that bad, but we do need to tighten up the quoteless path.

`VERSION 1.x:y` would match as `VERSION 1.x` followed by PNAME `:y`, as `:` is not a component of VERSION_SPECIFIER. But, if a parser does a blind tokenization, it will need to account for the possible matches based on context. The only token to match after `VERSION` (or `@version`) is `VERSION_SPECIFIER`. So, even though `1.x` could be tokenized as an integer `1` followed by a full-stop, that particular set of tokens cannot follow `VERSION`. (Similar mis-matches could come with matching `prefix` or `base` or a `PNAME`),

There there is a problem would be for something like `VERSION 1.x` followed by `y:z a rdfs:Resource` in the case where spaces are removed, it could parse as version `1.x` or version `1.xy`. (We faced similar ambiguities in the past with DECIMAL, which was previously `[0-9]+ "." [0-9]* | "." [0-9]+` which meant that the trailing '.' could be parsed as part of the decimal value, or it could be parsed as an INTEGER followed by full-stop. The addition of whitespace disambiguates, but we typically don't like to depend on whitespace to separate tokens.

I think the best change is to change the terminal to `([0-9] ([.-] [0-9]+]*) | ('"' [0-9a-z]+ ([.-] [0-9a-z]+)* '"')`. This would allow `VERSION 1.2345`, but not `VERSION 1.2x`, the latter would require the quoted form.

```suggestion
VERSION_SPECIFIER ::= ([0-9]+ ([.-] [0-9]+)*)
```

spec/turtle-bnf.html
```diff
@@ -379,53 +391,59 @@ <h3 id="terminals">Productions for terminals</h3>
...
...-z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">.-</code><code class="grammar-brac">]</code> <code class="grammar-brac">[</code><code class="grammar-literal">0-9a-z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code><code class="grammar-paren">)</code><code class="grammar-star">*</code><code class="grammar-paren">)</code> <code class="grammar-alt">|</code> <code class="grammar-paren">(</code>'<code class="grammar-literal">&quot;</code>' <code class="grammar-brac">[</code><code class="grammar-literal">0-9a-z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">.-</code><code class="grammar-brac">]</code> <code class="grammar-brac">[</code><code class="grammar-literal">0-9a-z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code><code class="grammar-paren">)</code><code class="grammar-star">*</code> '<code class="grammar-literal">&quot;</code>'<code class="grammar-paren">)</code></td>
```
gkellogg, line 445:
```suggestion
      <td><code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">.-</code><code class="grammar-brac">]</code> <code class="grammar-brac">[</code><code class="grammar-literal">0-9</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code><code class="grammar-paren">)</code><code class="grammar-star">*</code><code class="grammar-paren">)</code> <code class="grammar-alt">|</code> <code class="grammar-paren">(</code>'<code class="grammar-literal">&quot;</code>' <code class="grammar-brac">[</code><code class="grammar-literal">0-9a-z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code> <code class="grammar-paren">(</code><code class="grammar-brac">[</code><code class="grammar-literal">.-</code><code class="grammar-brac">]</code> <code class="grammar-brac">[</code><code class="grammar-literal">0-9a-z</code><code class="grammar-brac">]</code><code class="grammar-plus">+</code><code class="grammar-paren">)</code><code class="grammar-star">*</code> '<code class="grammar-literal">&quot;</code>'<code class="grammar-paren">)</code></td>
```

spec/turtle.bnf
kasei: Was there a discussion somewhere on the value of this quote-less form? It seems to me that the only value is in saving two bytes/keystrokes for something that in most cases will appear at most once in a file. That seems like a pretty small benefit, especially if we're introducing new terminals into the grammar that conflict with already existing ones. If we're set on the quote-less form, why not just use an existing terminal like `DECIMAL`?

ektrah: Being able to tokenize blindly/in a context-free manner is immensely useful, for example, when performing syntax highlighting or incremental parsing. Requiring context-sensitive tokenization would complicate things a lot.

I agree with https://github.com/w3c/rdf-turtle/pull/90#discussion_r2049200509 - don't complicate things and leave semantic checks on the version specifier to higher-level code:

```suggestion
VERSION_SPECIFIER ::= DECIMAL | STRING_LITERAL_QUOTE
```

gkellogg: > Was there a discussion somewhere on the value of this quote-less form? It seems to me that the only value is in saving two bytes/keystrokes for something that in most cases will appear at most once in a file. That seems like a pretty small benefit, especially if we're introducing new terminals into the grammar that conflict with already existing ones. If we're set on the quote-less form, why not just use an existing terminal like `DECIMAL`?

It was requested, I think by @rubensworks, but the value may not be e worth it.

DECIMAL would allow a number of unintended matches.

domel: +1 for STRING_LITERAL_QUOT and - 1 for DECIMAL

rubensworks: The short-hand was one of the suggestions in https://github.com/w3c/rdf-star-wg/issues/141.
But given the parsing complexities, it may not be worth it indeed.
I'm ok with only having the quoted string form.

domel: I think that it can be good do semi implement Semantic Versioning https://semver.org/. I mean MAJOR and MINOR.

Tpt: > It was requested, I think by @rubensworks, but the value may not be worth it.

I think I am the person who suggested the current grammar.

I agree that there are too many ambiguities and sorry for that. Something like `[0-9]+ ([.-] [0-9]+)*` or maybe even `[0-9]+ ([.-] [0-9]+)?` (to match only MAJOR or MAJOR.MINOR) seems much safer and will covers the most likely common use cases (and we can always extend it in a backward-compatible whay in the future

afs: The quoteless form can not be made to work with content-sensitive tokenization.

`DECIMAL` can start `+` or `-`.
`DECIMAL` does not allow `1.2.3`.

Plain 1.2 matches both VERSION and DECIMAL. A tokenizer without context will return one, not the other (and in javacc, it is a silent choice - the rule is "first token definition" so one of DECIMAL and VERSION will be returned breaking the other usage,

[Kasei's sugegstion](https://github.com/w3c/rdf-turtle/pull/90/files#r2049200509) of string only, + validation checks, looks like the only way within the current assumptions of parser complexity.

The string form should be both string forms `"1.2"` or `'1.2'`.

There is also a tokenizing ambiguity with `@version` and it is, technically, breaking old code (!) -- a langtag of `@version`. Parsers must currently be working around this for `@prefix` and `@base` but data could, theoretically, have such a language tag.

spec/index.html
```diff
@@ -299,6 +299,74 @@ <h3>Object Lists</h3>
     Turtle provides a number of ways of writing each.
   </p>
 
+  <section id="sec-version">
+    <h3>Version Announcement</h3>
+
+    <p>The Turtle language has evolved since its origin, and RDF 1.2
+      introduces new features which may confuse non-compliant parsers.
```
afs, line 306:
```suggestion
      adds new syntax.
```

spec/index.html
```diff
@@ -299,6 +299,74 @@ <h3>Object Lists</h3>
...
+      RDF 1.2 Turtle introduces the
+      <a href="#grammar-production-sparqlVersion"><code>VERSION</code></a> and
+      <a href="#grammar-production-version"><code>@version</code></a> directives
+      along with an optional `version` <a href="#sec-mediaReg">Media Type parameter</a>.
+      When serializing Turtle with new features such as base direction or triple terms,
+      authors can announce the presense of these features using these directives.
```
afs, line 312:
```suggestion
      authors can announce the use of the new syntax forms using these directives.
```

spec/index.html
```diff
@@ -299,6 +299,74 @@ <h3>Object Lists</h3>
...
+    </p>
+
+    <pre id="ex-version-decl"
+         class="example turtle" data-transform="updateExample"
+         title="Version announcement">
+      <!--
+      @version "1.2" .
+      @prefix : <http://example.com/> .
```
afs, line 320:
```suggestion
      VERSION "1.2"
      PREFIX : <http://example.com/>
```
(We have been changing examples to use "new" syntax so I think this should go first.

spec/index.html
```diff
@@ -299,6 +299,74 @@ <h3>Object Lists</h3>
...
+
+    <p>Alternatively, the SPARQL style directive can be used
+      (note that there is no trailing <a href="#cp-full-stop"><code title="full stop">.</code></a>):</p>
+
+    <pre id="ex-sparql-version-decl"
+         class="example turtle" data-transform="updateExample"
+         title="SPARQL-style version announcement">
+      <!--
+      VERSION "1.2"
+      PREFIX : <http://example.com/>
```
afs, line 333:
```suggestion
      @version "1.2" .
      @prefix : <http://example.com/> .
```

spec/index.html
```diff
@@ -299,6 +299,74 @@ <h3>Object Lists</h3>
...
+      :a :name "Alice" ~ :t {| :statedBy :bob ; :recorded "2021-07-07"^^xsd:date |} .
+      -->
+    </pre>
+
+    <p>The surrounding
+      quotation marks (<a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>)
+      may also be omitted.</p>
```
afs, line 340: Revise in light of other comments.

spec/index.html
```diff
@@ -476,6 +549,16 @@ <h3>Quoted Literals</h3>
         and may be followed by a <a data-cite="RDF12-CONCEPTS#dfn-base-direction">base direction</a>
         separated from the language tag by <a href="#cp-hyphen-hyphen"><code>--</code></a>.</p>
 
+      <p class="note">When using RDF 1.2-specific features, such as
+        <a data-cite="RDF12-CONCEPTS#dfn-base-direction">base direction</a>,
```
afs, line 553:
```suggestion
        <a data-cite="RDF12-CONCEPTS#dfn-base-direction">initial text direction</a>,
```

afs: @gkellogg - the link and text in RDF Concept is "base direction" then it says "initial text direction".

spec/index.html
```diff
@@ -798,9 +882,20 @@ <h3>Triple Terms</h3>
...
      </p>
 
+    <p class="note">When using RDF 1.2-specific features, such as
+      <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>,
+      the specific RDF version should be announced using
+      either the <a href="#grammar-production-sparqlVersion"><code>VERSION</code></a>
+      or <a href="#grammar-production-version"><code>@version</code></a>
+      directives early in the document.
+      This allows parsers that do not support these features to detect
+      the presense of such features early.
```
afs, line 892:
```suggestion
      the presence of such features early.
```

spec/index.html
```diff
@@ -869,10 +964,21 @@ <h3>Reifying Triples</h3>
...
       and used to identify this relationship.</p>
 
+    <p class="note">When using RDF 1.2-specific features, such as
+      <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>,
+      the specific RDF version should be announced using
+      either the <a href="#grammar-production-sparqlVersion"><code>VERSION</code></a>
+      or <a href="#grammar-production-version"><code>@version</code></a>
+      directives early in the document.
+      This allows parsers that do not support these features to detect
+      the presense of such features early.
```
afs, line 974:
```suggestion
      the presence of such features early.
```

spec/index.html
```diff
@@ -954,10 +1060,21 @@ <h2>Annotation Syntax</h2>
         and the RDF Abstract Syntax [[RDF11-CONCEPTS]] does not
         distinguish how the triples were written.</p>
 
+      <p class="note">When using RDF 1.2-specific features, such as
```
afs, line 1063: This seems to be duplicate text to lines 967-974.

spec/turtle.bnf
gkellogg: > The string form should be both string forms `"1.2"` or `'1.2'`.

To save current tokenizers, as suggested by @ektrah, probably best to use the following rule:

```bnf
VERSION_SPECIFIER ::= STRING_LITERAL_QUOTE | `STRING_LITERAL_SINGLE_QUOTE
```

We already say that anything other than `"1.2"` is not defined.

> There is also a tokenizing ambiguity with `@version` -- theoretically, RDF 1.1 data may have a langtag of `@version`. Parsers must currently be working around this for `@prefix` and `@base` - depending on how they do that, it might have an impact.

Although `@version` matches `LANG_DIR` (as does `@base` and `@prefix`), RDF Concepts requires that language tags be well-formed according to BCP47, which none of these are. (Perhaps Turtle should call this out, as well). Even though parsers may treat `@version` as a language tag, it would always result in invalid RDF. Probably worth noting in the changes section, though.

spec/index.html
gkellogg: This example mirrors the [`@prefix` example](https://www.w3.org/TR/rdf12-turtle/#ex-prefix-decl) in [2.5 IRI](https://www.w3.org/TR/rdf12-turtle/#sec-iri). We should have at least one example using `@version`.

spec/index.html
gkellogg: If you're suggesting that the SPARQL-style `VERSION` example should show before the old-style `@version` example, then this is at odds with the example ordering in the IRI section.

spec/turtle.bnf
afs: I'm not worried by this because they are reserved (4 alpha) or the `5*8` rule.

They are allowed in RFC 5646/BCP 47. In [RFC 4646](https://datatracker.ietf.org/doc/html/rfc4646#section-2.2.1) there is a discussion about the possibility but that text is not in RFC 5654

```
 language      = 2*3ALPHA            ; shortest ISO 639 code
                 ["-" extlang]       ; sometimes followed by
                                     ; extended language subtags
               / 4ALPHA              ; or reserved for future use
               / 5*8ALPHA            ; or registered language subtag
```

afs: > `VERSION_SPECIFIER ::= STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE`

+1
possibly together with @kasei's suggestion of a semantic check, or stick with the only "1.2" defined for now.

spec/index.html
afs: +1

Its copied verbatim so the annotation syntax refers to triple terms which aren't visible.

spec/turtle.bnf
gkellogg: > The "must be valid" approach does not work because it is backtracking (i.e. changing the token decision) after accepting a token.

This is still a requirement of parsers, even if the tokenizer treats `@base`, `@prefix`, or `@version` as a language tag. Presumably, 1.1 parsers prioritize matching the declarations before `LANG_DIR` (really `LANGTAG` then). A 1.1 parser may treat `@version` as a language tag, but will have an error when it is followed by a `VERSION_SPECIFIER`, which accomplishes the same thing.

> possibly together with @kasei's suggestion of a semantic check, or stick with the only "1.2" defined for now.

Until we come up with some other values, saying that values other than "1.2" don't have a defined behavior should be enough.

spec/index.html
gkellogg: It's in the commits.

spec/index.html
gkellogg: Now it should only appear in the "Version Announcement" section.

spec/index.html
```diff
@@ -308,43 +308,31 @@ <h3>Version Announcement</h3>
       <a href="#grammar-production-sparqlVersion"><code>VERSION</code></a> and
       <a href="#grammar-production-version"><code>@version</code></a> directives
       along with an optional `version` <a href="#sec-mediaReg">Media Type parameter</a>.
-      When serializing Turtle with new features such as base direction or triple terms,
+      When serializing Turtle with new features such as
+      <a data-cite="RDF12-CONCEPTS#dfn-base-direction">initial text direction</a>
+      or<a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>,
```
TallTed, line 313:
```suggestion
      or <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>,
```

spec/index.html
```diff
@@ -362,6 +350,16 @@ <h3>Version Announcement</h3>
       -->
     </pre>
 
+    <p class="note">When using RDF 1.2-specific features, such as
+      <a data-cite="RDF12-CONCEPTS#dfn-base-direction">initial text direction</a>,
+      the specific RDF version should be announced using
+      either the <a href="#grammar-production-sparqlVersion"><code>VERSION</code></a>
```
TallTed, line 356:
```suggestion
      either <a href="#grammar-production-sparqlVersion"><code>VERSION</code></a>
```

spec/index.html
```diff
@@ -362,6 +350,16 @@ <h3>Version Announcement</h3>
...
+      or <a href="#grammar-production-version"><code>@version</code></a>
+      directives early in the document.
+      This allows parsers that do not support these features to detect
+      the presense of such features early.
```
TallTed, line 360:
```suggestion
      the presense of such features early, and therefore 
      cease processing a job that would ultimately fail.
```

spec/index.html
```diff
@@ -2360,16 +2318,19 @@ <h2>Changes between RDF 1.1 and RDF 1.2</h2>
...
       <a data-cite="I18N-GLOSSARY#dfn-surrogate" class="lint-ignore">Unicode surrogates</a> 
       are not legal in Turtle documents, nor as the value of a Unicode escape sequence.</li>
     <li>Added <a href="#sec-version"></a> and parser
-      state and productions to announce the RDF version associated with the input document.</li>
+      state and productions to announce the RDF version
+      associated with the input document.
+      Note that this also includes a provision that `@version` does not match
+      <a href="#grammar-production-LANG_DIR" class="type langDir"><code>LANG_DIR</code></a>;
+      the previous version of the grammar would allow such a match,
+      but would result in a literal with a non-well-formed language tag accroding to [[BCP47]], and form an invalid literal.</li>
```
TallTed, line 2331: `non-well-formed` feels clumsy. BCP47 (consisting of [RFC 4647](https://www.rfc-editor.org/info/rfc4647) and [RFC 5646](https://www.rfc-editor.org/info/rfc5646)) does not contain `non-well-formed` anywhere, but it *does* have one instance of `ill-formed` (and 17 instances of `well-formed). Can we use `ill-formed` instead of `non-well-formed`?
```suggestion
      but would result in a literal with an ill-formed language tag according to [[BCP47]], and form an invalid literal.</li>
```

spec/turtle.bnf
kasei: > possibly together with @kasei's suggestion of a semantic check, or stick with the only "1.2" defined for now.

My intention in recommending that was mostly for implementors if they cared about the only accepting recognized versions. I don't think the spec should deal with the higher level checking.

spec/index.html
```diff
@@ -308,43 +308,31 @@ <h3>Version Announcement</h3>
...
       PREFIX : <http://example.com/>
       :a :name "Alice" ~ :t {| :statedBy :bob ; :recorded "2021-07-07"^^xsd:date |} .
       -->
     </pre>
 
-    <p>The surrounding
-      quotation marks (<a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>)
-      may also be omitted.</p>
+    <p>Alternatively, the old-style directive can be used
+      (note that there is no trailing <a href="#cp-full-stop"><code title="full stop">.</code></a>):</p>
```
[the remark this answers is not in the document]
gkellogg:
```suggestion
      (note that there is no trailing &quot;<a href="#cp-full-stop"><code title="full stop">.</code></a>&quot;):</p>
```

spec/index.html
gkellogg: I'm not that happy with the new proposed wording. Perhaps: "and potentially halt processing early."

TallTed: Maybe?
```suggestion
      the presense of such features early, and potentially 
      inform the user, giving them an opportunity to stop the job
      or otherwise act on the fact that some amount of the input data 
      will not be processed as desired.
```

spec/turtle.bnf
```diff
@@ -81,3 +83,4 @@ PLX               ::= PERCENT | PN_LOCAL_ESC
 PERCENT           ::= '%' HEX HEX
 HEX               ::= ([0-9] | [A-F] | [a-f])
 PN_LOCAL_ESC      ::= '\' ( '_' | '~' | '.' | '-' | '!' | '$' | '&' | "'" | '(' | ')' | '*' | '+' | ',' | ';' | '=' | '/' | '?' | '#' | '@' | '%' )
+VERSION_SPECIFIER ::= STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE
```
ektrah, line 86: Is not allowing long quotation marks purely stylistic or is there a technical reason for it? I.e., if someone really wants to write `@version """1.2""" .`, is there any reason to stop them? If not, I'd propose allowing all syntactic forms of strings as this simplifies parsing slightly (just reuse the parser and AST definitions for string literals).

```suggestion
VERSION_SPECIFIER ::= STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE
                    | STRING_LITERAL_LONG_SINGLE_QUOTE | STRING_LITERAL_LONG_QUOTE
```

afs: We can't have `VERSION_SPECIFIER` as a token because it will match `STRING_LITERAL_QUOTE`.
That breaks the grammar rule [`String`](https://www.w3.org/TR/rdf12-turtle/#grammar-production-String). 

`STRING_LITERAL_QUOTE` will never be returned by a tokenizer - `VERSION_SPECIFIER`will be returned (some tokenizer will make it an error because of the ambiguity; javacc issues a warning).

The grammar rule `String` is then broken.

```
sparqlVersion     ::= "VERSION" ( STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE
                    | STRING_LITERAL_LONG_SINGLE_QUOTE | STRING_LITERAL_LONG_QUOTE )
```

as that's the [grammar rule for String](https://www.w3.org/TR/rdf12-turtle/#grammar-production-String) it can be written:
```
sparqlVersion         ::=	"VERSION" String
```

For single quotes only, it would be 

```
sparqlVersion         ::=	"VERSION" ( STRING_LITERAL_QUOTE | STRING_LITERAL_SINGLE_QUOTE)
```

now determination of the AST is in the grammar, and it's seen "VERSION" so it can look for specific tokens.
There is only one token choice for a given input sequence `"1.2"`.

(I'm experimenting with the javacc-based Turtle parser without any javacc specific features - this is plain LL(1) and so also LALR(1))

afs: I don't see the need to allow triple quoted forms, though they don't do any harm.

One minor disadvantage is that grep'ing a Turtle file for `VERSION` is a bit more tricky, given a convention of same line for `VERSION` (c.f. `PREFIX`) but this is not a major downside.

Personally, I'd not include them so as to make it less convenience to have newlines in the version string.

It make next to no difference in implementation effort.

spec/index.html
```diff
@@ -299,6 +299,75 @@ <h3>Object Lists</h3>
...
+      <!--
+      GET /document.ttl HTTP/1.1
+      Host: example.com
+      Accept: text/turtle; version=1.2
+      -->
+    </pre>
+
+    <p class="note">When using RDF 1.2-specific features, such as
+      <a data-cite="RDF12-CONCEPTS#dfn-base-direction">initial text direction</a>,
+      the specific RDF version should be announced using
```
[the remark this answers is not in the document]
afs: It's in a `class="note"`, which isn't normative.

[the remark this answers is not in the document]
ektrah: A bit above the text says "authors can announce the use of the new syntax forms using these directives", so I guess it's a MAY (a.k.a. OPTIONAL). In this case:

* As a document generator, I wouldn't bother generating a version directive &ndash; it's optional. (If I want to accurately declare the lowest version required for the features actually used in the document, I'd have to first generate the whole document and then go back and inject the version directive somewhere in the middle or beginning. That's difficult if I'm generating the document into a stream without buffering the whole document first.)
* As a document parser, I'd just just issue a warning if I don't recognize the version specifier but then just continue parsing, hoping that the document doesn't actually use any of the features I don't understand. (E.g., if the document declares a hypothetical 1.3 version but only uses 1.2 features, I'd want to parse the document anyway).

Is this the way the version directive is intended?

afs (changes requested):
Can't use a new token for the version specifier.

spec/turtle.bnf
gkellogg: The point is to have a term constructor which can recognize a specific production from which to extract the version string. We can change VERSION_SPECIFIER to VersionSpecifier and make it a non-terminal.

I’m not inclined to add triple-quote versions at this point.

spec/index.html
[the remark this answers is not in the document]
gkellogg: The WG intention is to encourage, but not require a version announcement, typically early in the file. A streaming serializer may not be able to do this until it goes to serialize a triple containing a triple term, but can output it at that point. That’s why these directions are non-normative notes.

spec/turtle.bnf
afs: Checked - `VersionSpecifier` works.

spec/index.html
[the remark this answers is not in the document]
TallTed: `should` is never meant to be treated as a `MUST` (firm requirement) nor a `MAY` (firm optional), though it is sometimes meant to be treated as a `SHOULD` (strong encouragement). As this particular text is non-normative, no all-caps wording is appropriate, and the best we can hope for is the intended moderate encouragement of lowercase `should`.

afs:
Will escape sequences be allowed in a version string?

My reading of the current draft is "no"  because the outcome of VersionSpecifier is not a `String` (grammar rule). The escape sequences text says "Strings" (written String, in bold; s, not in bold; no link).

Places other than Turtle, it may be possible to have these characters. e.g. XML-based formats will do entity processing regardless; JSON escapes similarly. 

There is a restriction that it should be usable in HTTP `version=` and best if without being a quote-string.


https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.6
```
  parameter       = parameter-name "=" parameter-value
  parameter-name  = token
  parameter-value = ( token / quoted-string )
```
https://datatracker.ietf.org/doc/html/rfc9110#name-tokens
```
  token          = 1*tchar

  tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                 / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                 / DIGIT / ALPHA
                 ; any VCHAR, except delimiters
```

gkellogg:
Replacing of escape sequences is in the STRING_* Term Constructor, so this should handle the VersionSpecifier case.

We also need to update 6.2 Comments – "Comments in Turtle start with a [#](http://localhost:8000/rdf-turtle/spec/#cp-number-sign) outside an [IRIREF](http://localhost:8000/rdf-turtle/spec/#grammar-production-IRIREF) or [String](http://localhost:8000/rdf-turtle/spec/#grammar-production-String)" This would be true for VersionSpecifier.

Or, we could change this to "IRIREF, STRING_LITERAL_SINGLE_QUOTE, STRING_LITERAL_QUOTE, STRING_LITERAL_LONG_SINGLE_QUOTE, or STRING_LITERAL_LONG_QUOTE".

I think token version specifiers can be revisited in the future after there's time for feedback.

gkellogg:
@afs, needs your approval to allow merging.

spec/index.html
```diff
@@ -1876,7 +1880,7 @@ <h3>RDF Term Constructors</h3>
...
-      <tr id="handle-VersionSpecifier"                ><td style="text-align:left;"            ><a class="type literal"      href="#grammar-production-VersionSpecifier"                >VersionSpecifier                </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-literal">      literal.     </a></td><td>The |curVersion| is taken from a literal having a lexical form the matched value exclusive of surrounding <a href="#cp-quotation-mark"><code title="quotation mark">&quot;</code></a>s or <a href="#cp-apostrophe"><code title="apostrophe">'</code></a>s and datatype `xsd:string`.</td></tr>
+      <tr id="handle-VersionSpecifier"                ><td style="text-align:left;"            ><a class="type literal"      href="#grammar-production-VersionSpecifier"                >VersionSpecifier                </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-literal">      literal.     </a></td><td>The |curVersion| is taken from a literal using the matched <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a> lexical form and datatype `xsd:string`.</td></tr>
```
TallTed, line 1883:
```suggestion
      <tr id="handle-VersionSpecifier"                ><td style="text-align:left;"            ><a class="type literal"      href="#grammar-production-VersionSpecifier"                >VersionSpecifier                </a></td><td><a data-cite="RDF12-CONCEPTS#dfn-literal">      literal.     </a></td><td>The |curVersion| is taken from a literal using the matched <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a> lexical form and `xsd:string` datatype.</td></tr>
```

PR #120, opened 2025-12-19
Refer to RDF Concepts for version labels

afs:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/120.html" title="Last updated on Dec 19, 2025, 9:10 PM UTC (d5624ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/120/405add3...d5624ac.html" title="Last updated on Dec 19, 2025, 9:10 PM UTC (d5624ac)">Diff</a>

PR #122, opened 2026-01-08
add clarity on how to deal with 0-N version directives

### LINKED ISSUE #118: explain how to interpret multiple `VERSION` directives
In [§2.4 Version Announcement](https://www.w3.org/TR/rdf12-turtle/#sec-version), we should point out that multiple `VERSION` directives can co-exist in a file, that each of them is assumed to apply to the content after it and before the next `VERSION` directive (if any) -- keeping in mind that those directives are mere hints, and do not strongly constrain the syntax nor the parsing.

This should address part of the comments in https://github.com/w3ctag/design-reviews/issues/1161#issuecomment-3671223503
(since N-Triples refers to this section of Turtle).

pchampin:
fix #118

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/122.html" title="Last updated on Jan 15, 2026, 4:28 PM UTC (f85c57e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/122/a6c9960...f85c57e.html" title="Last updated on Jan 15, 2026, 4:28 PM UTC (f85c57e)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -425,13 +425,13 @@
       (but could signal them with a warning).
     </p>
 
-    <div class="note">Acceptable version numbers are the ones defined in
-      <a data-cite="RDF12-CONCEPTS#defined-version-labels">2.1 Version Labels</a>
-      in [[RDF12-CONCEPTS]].
-      Processors may treat other version numbers as an error or warning.
-
+    <div class="note">Version labels are defined in [[[RDF12-CONCEPTS]]] section
+      <a data-cite="RDF12-CONCEPTS#defined-version-labels">2.1 Version Labels</a>.
+      Processors may treat unrecognized labels as an error or as a warning.
       <p>
-        While `"1.1"` is an acceptable version label, its use in a `VERSION` or `@version` directive is discouraged, as it would needlessly cause Turtle 1.1 parsers to fail.
+        While `"1.1"` is an acceptable version label,
+        its use in a `VERSION` or `@version` directive is discouraged,
+        as it would needlessly cause Turtle 1.1 parsers to fail.
       </p>
     </div>
   </section>
--- a/spec/index.html
+++ b/spec/index.html
@@ -412,17 +412,22 @@
       will not be processed as desired.
     </p>
 
-    <p>Multiple version announcement directives may appear in a Turtle document, each of them applying to the part of the document following it and until the next one or the end of the document.
+    <p>Multiple version directives may appear in a Turtle document.
+    The directive applies to the part of the document following the directive,
+    until another directive is encountered or the end of the document is reached.
     </p>
 
     <p>
-     In the absence of a version announcement directive, the version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered, if any.
+     In the absence of a current version directive, any version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered.
      In the absence of any version information, a sensible default is `"1.1"`, to maximize compatibility with the previous version of Turtle [[TURTLE]].
       In any case, the version announcement is merely a hint;
-      parsers are not required to reject features that are outside the announced version (but could signal them with a warning).
+      parsers are not required to reject features that are outside the announced version
+      (but could signal them with a warning).
     </p>
 
-    <div class="note">Acceptable version numbers are the ones defined in <a data-cite="RDF12-CONCEPTS#defined-version-labels">2.1 Version Labels</a> in [[RDF12-CONCEPTS]].
+    <div class="note">Acceptable version numbers are the ones defined in
+      <a data-cite="RDF12-CONCEPTS#defined-version-labels">2.1 Version Labels</a>
+      in [[RDF12-CONCEPTS]].
       Processors may treat other version numbers as an error or warning.
 
       <p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -399,7 +399,7 @@
       -->
     </pre>
 
-    <p class="note">When using RDF 1.2-specific features, such as
+    <p>When using RDF 1.2-specific features, such as
       <a data-cite="RDF12-CONCEPTS#dfn-base-direction">initial text direction</a>,
       the specific RDF version should be announced using
       either <a href="#grammar-production-sparqlVersion"><code>VERSION</code></a>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -412,9 +412,23 @@ <h3>Version Announcement</h3>
...
 
-    <p class="note">Version numbers other than "1.2" are not defined by
-      this specification.
-      Processors may treat this as an error or warning.</p>
+    <p>Multiple version announcement directives may appear in a Turtle document, each of them applying to the part of the document following it and until the next one or the end of the document.
+    </p>
+
+    <p>
+     In the absence of a version announcement directive, the version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered, if any.
+     In the absence of any version information, a sensible default is `"1.1"`, to maximize compatibility with the previous version of Turtle [[TURTLE]].
```
Tpt, line 422: If this is done by a parser, it might lead the parser to fail on triple terms. And similarly if this is done by a serializer. This default seems pointing implementation to effectively require `VERSION 1.2` to be there if 1.2 features are used.

spec/index.html
```diff
@@ -412,9 +412,23 @@ <h3>Version Announcement</h3>
       will not be processed as desired.
     </p>
 
-    <p class="note">Version numbers other than "1.2" are not defined by
-      this specification.
-      Processors may treat this as an error or warning.</p>
+    <p>Multiple version announcement directives may appear in a Turtle document, each of them applying to the part of the document following it and until the next one or the end of the document.
```
afs:
```suggestion
    <p>Multiple version directives may appear in a Turtle document.
    The directive applies to the part of the document following the directive,
    until another directive is encountered or the end of the document is reached.
```

spec/index.html
```diff
@@ -412,9 +412,23 @@ <h3>Version Announcement</h3>
...
+    </p>
+
+    <p>
+     In the absence of a version announcement directive, the version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered, if any.
```
afs: > In the absence of a version announcement directive, 

That reads as requiring the whole document to be processed before deciding to use the version parameter.

It needs a sense of "version currently in force" - this is what the parser state does.

Maybe:

```suggestion
     In the absence of current a version directive, the version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered, if any.
```
but more wording may better.

spec/index.html
```diff
@@ -412,9 +412,23 @@ <h3>Version Announcement</h3>
...
+     In the absence of any version information, a sensible default is `"1.1"`, to maximize compatibility with the previous version of Turtle [[TURTLE]].
+      In any case, the version announcement is merely a hint;
+      parsers are not required to reject features that are outside the announced version (but could signal them with a warning).
+    </p>
+
+    <div class="note">Acceptable version numbers are the ones defined in <a data-cite="RDF12-CONCEPTS#defined-version-labels">2.1 Version Labels</a> in [[RDF12-CONCEPTS]].
```
afs: Can we have short lines please?

afs (changes requested):

spec/index.html
afs: See #120.

It refers to RDF Concepts as the place for versions.

```
 <p class="note">Version labels are defined in [[[RDF12-CONCEPTS]]] section
      <a data-cite="RDF12-CONCEPTS#defined-version-labels">2.4 Version Announcement</a>.
      Processors may treat unrecognized labels as an error or as a warning.
    </p>
```

See also https://github.com/w3c/rdf-concepts/pull/261 which adds text about how "1.1" is compatible with "1.2".

spec/index.html
TallTed:
```suggestion
     In the absence of a current version directive, any version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered.
```

pchampin: well spotted, thanks.

spec/index.html
pchampin: will do

pchampin:
```suggestion
      parsers are not required to reject features that are outside the announced version
      (but could signal them with a warning).
    </p>

    <div class="note">Acceptable version numbers are the ones defined in
      <a data-cite="RDF12-CONCEPTS#defined-version-labels">2.1 Version Labels</a>
      in [[RDF12-CONCEPTS]].
```

spec/index.html
pchampin: sorry I overlooked #120

I'll consolidate it into this PR, as they will otherwise conflict.

pchampin: e60be6f reuses the text in #120.

spec/index.html
pchampin: > This default seems pointing implementation to effectively require VERSION 1.2 to be there if 1.2 features are used.

No, not "require". The next sentence says:

"In any case, the version announcement is merely a hint; parsers are not required to reject features that are outside the announced version (but could signal them with a warning)."

So yes, the encouraged practice would be:
* old Turtle file or 1.1-compatible file contain no VERSION directive
* files with Triple-terms contain VERSION "1.2", so that old parser can err early, with an explicit message
but lenient parsers can still accept triple terms if no VERSION directive is given.

domel (approved):

afs (approved):

afs (approved):

spec/index.html
Tpt: Thank you!

> So yes, the encouraged practice would be:

Agreed to the practice for serializers. However, I do not see how it relates to my comment that was aimed at parsers. In this case, just pushing them to do even things like "signaling as a warning" in case of absence of version information looks quite cumbersome to me in real life. I would tend to encourage as broad compatibility as possible.

pchampin: > In this case, just pushing them to do even things like "signaling as a warning" in case of absence of version information looks quite cumbersome to me in real life. I would tend to encourage as broad compatibility as possible.

+1 to encourage broad compatibility; that is why I would discourage parsers to raise errors when the content of a file does not match the announced version.

But I disagree that raising a warning (when RDF 1.2 features are encountered without proper version announcement) is hindering compatibility. It is a *only* warning after all...

That being said, I grant you that the warning may not reach the right target anyway... If the server does not follow good practices and sends triple terms without any version announcement, it is the client who will get a warning when parsing that...

The PR has been merged, but feel free to make a counter-proposal as a new PR if you think a better wording is required.

PR #135, opened 2026-04-20
Add content negotiation subsection

niklasl:
This does two things:

1. Fixes the incorrect example about version announcement, which currently uses the `Accept` HTTP *request* header as an example of a server announcing the version of a turtle document being served. That is a *response*, and needs to use the `Content-Type` HTTP *response* header to provide information about the media type being served.

2. Adds information about syntax versioning and content negotiation on the web. Should this rather go elsewhere; e.g. in RDF Concepts? It is rather technical, and probably better suited in a concrete syntax document. On the other hand, this same reasoning goes for the other syntaxes as well. Is it OK for them to link to this section in the Turtle spec?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/135.html" title="Last updated on Apr 22, 2026, 8:26 AM UTC (3b35ce9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/135/b4dd46c...3b35ce9.html" title="Last updated on Apr 22, 2026, 8:26 AM UTC (3b35ce9)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -435,6 +435,60 @@
       </p>
     </div>
 
+    <section id="sec-version-conneg">
+      <h4>Syntax Version and Content Negotiation</h4>
+
+      <p>Just as HTTP servers can announce the syntax version of Turtle content
+        provided, HTTP clients can use the `version` media type parameter in
+        the `Accept` header, for the purpose of
+        <a data-cite="RFC9110#section-12.5.1">Content Negotiation</a>,
+        as follows:</p>
+      <pre id="ex-http-version-request"
+          class="example http" data-transform="updateExample"
+          title="HTTP content negotation for Turtle using version parameter">
+        <!--
+        GET /document HTTP/1.1
+        Host: example.com
+        Accept: text/turtle; version=1.2
+        -->
+      </pre>
+
+      <p>Note that the accept header of a client does not have to make use of
+        the `version` media type parameter. A request can simply look like:</p>
+      <pre id="ex-http-versionless-request"
+          class="example http" data-transform="updateExample"
+          title="HTTP content negotation for Turtle without version parameter">
+        <!--
+        GET /document HTTP/1.1
+        Host: example.com
+        Accept: text/turtle
+        -->
+      </pre>
+      <p>This carries no implication about the parsing capabilities of the
+        client. However, since the client doesn't advertise what version of
+        Turtle it is capable of parsing, a server might be conservative
+        here and explicitly serve 1.1 content instead:</p>
+      <pre id="ex-http-version-11-response"
+          class="example http" data-transform="updateExample"
+          title="HTTP response announcing Turtle version 1.1 response">
+        <!--
+        HTTP/1.1 200 OK
+        Content-Type: text/turtle; version=1.1
+        Content-Location: https://example.com/document-basic.ttl
+        -->
+      </pre>
+      <p>Note that while content negotiation follows clear rules, it is a "best
+        available" design, and a server is free to simply serve what it has,
+        without regard for the version parameter. The specific circumstances
+        leading to a conservative response, as above, are beyond the scope of
+        this document to define. Other services might, for example,
+        conditionally serve different syntaxes when the `version` parameter is
+        used in combination with the `q`
+        <a data-cite="RFC9110#section-12.4.2">media type preference</a>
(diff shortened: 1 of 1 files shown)


domel:
I see three substantive issues here.

First, the default version is inconsistent with RDF 1.2 Concepts. Section 2.4 says that, in the absence of any version information, a sensible default is "1.1", while RDF 1.2 Concepts allows processors to assume "1.2" when no version is announced. This should be aligned.

Second, the precedence rule is unclear. The current wording suggests that the media type parameter is only considered when there is no active `VERSION` / `@version` directive. However, RDF 1.2 Concepts says that when both are present and disagree, the media type parameter takes precedence, optionally with a warning. That should be stated explicitly.

Third, in 2.4.1, the `Accept: text/turtle` example should be phrased more carefully. Without a `version` parameter, it only indicates a preference for Turtle in general, not specifically 1.1. Serving 1.1 is a possible server policy, but it should not read as the default or implied outcome of that header.

afs:
Please split into two PRs so we can apply the fix ASAP.

afs:
> 2\. Adds information about syntax versioning and content negotiation on the web. Should this rather go elsewhere; e.g. in RDF Concepts? It is rather technical, and probably better suited in a concrete syntax document. On the other hand, this same reasoning goes for the other syntaxes as well. Is it OK for them to link to this section in the Turtle spec?

The text should be non-normative; it's explaining, not defining.

An option is an appendix in RDF Concepts. "version" is general and new; it applies to RDF/XML as well as the SPARQL protocol.

niklasl:
> Please split into two PRs so we can apply the fix ASAP.

@afs Good call; created https://github.com/w3c/rdf-turtle/pull/136.

niklasl:
@domel

> I see three substantive issues here.
> 
> First, the default version is inconsistent with RDF 1.2 Concepts. Section 2.4 says that, in the absence of any version information, a sensible default is "1.1", while RDF 1.2 Concepts allows processors to assume "1.2" when no version is announced. This should be aligned.

I agree. And as @afs says, this might also be better as an appendix in Concepts.

I see that [section 1.10 in Concepts](https://www.w3.org/TR/rdf12-concepts/#section-version-announcement) may be the more appropriate place to make this addition. I also see that the `Location` header in that example needs correcting too! (That header is used for [redirects](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Location).)
 
> Second, the precedence rule is unclear. The current wording suggests that the media type parameter is only considered when there is no active `VERSION` / `@version` directive. However, RDF 1.2 Concepts says that when both are present and disagree, the media type parameter takes precedence, optionally with a warning. That should be stated explicitly.

Good point. This proposed addition currently only focuses on announcement and negotiation, not on consolidating conflicting advice on how to deal with opposing signals in the media type version and actual content.

Do we need a coordination issue to decide how to align this? Otherwise, if we agree that this should go in Concepts, I'll close this (the required adjustment is in #136), and make some time to make a new PR on Concepts where we can coordinate the change.
 
> Third, in 2.4.1, the `Accept: text/turtle` example should be phrased more carefully. Without a `version` parameter, it only indicates a preference for Turtle in general, not specifically 1.1. Serving 1.1 is a possible server policy, but it should not read as the default or implied outcome of that header.

I agree that this needs careful explanation. I tried to point this out, but I'll think of more effective wording  (suggestions always appreciated of course).

afs:
> Do we need a coordination issue to decide how to align this?

Yes - and some telcon time.

> I see that [section 1.10 in Concepts](https://www.w3.org/TR/rdf12-concepts/#section-version-announcement) may be the more appropriate place to make this addition.

Section 1 is "Introduction" and 1.10 is already a bit towards "not introduction" to me. Yes, talk about version, but a significant amount of content on protocol is getting into too deep at this point, RDF is data -- my concern here is this is creeping towards a section on "RDF Protocol".

As a result of the TAG discussion, we may want general material covers error handling to make it clear that RDF 1.2 features cause parse errors.

afs:
Title changed to reflect #136 being separated out.

niklasl:
Created https://github.com/w3c/rdf-star-wg/issues/189 as a tracking issue, and added some of the above points (please edit the description if it is unclear).

niklasl:
Closing this; changes will probably go in Concepts (see previous comment).

PR #136, opened 2026-04-20
Fix version HTTP response example

niklasl:
This fixes the incorrect example about version announcement, which currently uses the `Accept` HTTP *request* header as an example of a server announcing the version of a turtle document being served. That is a *response*, and needs to use the `Content-Type` HTTP response header to provide information about the media type being served.

(This is the fix from https://github.com/w3c/rdf-turtle/pull/135 which also proposes additional text about content negotiation.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/136.html" title="Last updated on Apr 20, 2026, 7:53 PM UTC (52eee8b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/136/b4dd46c...52eee8b.html" title="Last updated on Apr 20, 2026, 7:53 PM UTC (52eee8b)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -395,7 +395,6 @@
       <!--
       HTTP/1.1 200 OK
       Content-Type: text/turtle; version=1.2
-      Content-Location: https://example.com/document.ttl
       -->
     </pre>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -389,13 +389,13 @@
     <p>When providing content over HTTP, servers can announce the version
       using the optional `version` <a href="#sec-mediaReg">Media Type parameter</a>:</p>
 
-    <pre id="ex-http-version-decl"
-         class="example turtle" data-transform="updateExample"
-         title="HTTP version announcement">
+    <pre id="ex-http-version-12-response"
+         class="example http" data-transform="updateExample"
+         title="HTTP response announcing Turtle version 1.2 response">
       <!--
-      GET /document.ttl HTTP/1.1
-      Host: example.com
-      Accept: text/turtle; version=1.2
+      HTTP/1.1 200 OK
+      Content-Type: text/turtle; version=1.2
+      Content-Location: https://example.com/document.ttl
       -->
     </pre>
 


spec/index.html
```diff
@@ -389,13 +389,13 @@ <h3>Version Announcement</h3>
...
+    <pre id="ex-http-version-12-response"
+         class="example http" data-transform="updateExample"
+         title="HTTP response announcing Turtle version 1.2 response">
       <!--
-      GET /document.ttl HTTP/1.1
-      Host: example.com
-      Accept: text/turtle; version=1.2
+      HTTP/1.1 200 OK
+      Content-Type: text/turtle; version=1.2
+      Content-Location: https://example.com/document.ttl
```
Tpt: do we need this line? It confused me a bit

niklasl: @tpt This one can be removed if preferred:
```suggestion
```
(I kept it from the `GET` and since the systems I mainly use has that in their response.)

niklasl: It does depend on whether we want this as part of a more cohesive example about content negotiation, where `Content-Location` is [commonly used](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Location) to provide a direct URL to the actual representation served.

Tpt: Yes! My feeling is that it's a bit confusing here because we are only describing a specific content negotiation feature and not everything. But it's really a nitpick, I am fine with the MR as it is.

niklasl: I share you concern; it made more sense (to me) as part of the narrative I suggest in #135. That might be better as part of Concepts; and a subsequent PR to Turtle could refer to that section, once/if we have it there. I'll remove the line here for now.

domel (approved):

pchampin (approved):

afs (approved):

spec/index.html
```diff
@@ -389,13 +389,12 @@ <h3>Version Announcement</h3>
     <p>When providing content over HTTP, servers can announce the version
       using the optional `version` <a href="#sec-mediaReg">Media Type parameter</a>:</p>
 
-    <pre id="ex-http-version-decl"
-         class="example turtle" data-transform="updateExample"
-         title="HTTP version announcement">
+    <pre id="ex-http-version-12-response"
+         class="example http" data-transform="updateExample"
+         title="HTTP response announcing Turtle version 1.2 response">
```
TallTed, line 394: The second "response" appears to be redundant?
```suggestion
         title="HTTP response announcing Turtle version 1.2">
```

PR #148, opened 2026-05-29
state that Turtle 1.2 is not ambiguous for 1.1 implems

pchampin:
also adapt the introductory example to include Turtle 1.2 features.

this PR is largely inspired by w3c/rdf-n-triples#90 and w3c/rdf-n-triples#105.

Fix #125

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/148.html" title="Last updated on Jun 4, 2026, 2:26 PM UTC (d39593d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/148/a84ef51...d39593d.html" title="Last updated on Jun 4, 2026, 2:26 PM UTC (d39593d)">Diff</a>

PR #150, opened 2026-05-29
Version fix

pchampin:
* [reword remark on unrecognized version](https://github.com/w3c/rdf-turtle/commit/7d9174f15556a1ff50b2ba10e8fad3f73f03abe1)

the previous wording seemed to imply that new features could be *misinterpreted*,
which is not the case.

* [remove confusing remark about 'sensible default'](https://github.com/w3c/rdf-turtle/commit/6ab938c0a57dc4c67b878f58e5a2b1359c1de5de)

This is indeed contradicting other pieces of advices in RDF Concept.
For servers, 1.1 is a sensible default in the absence of version in the Accept header,
but for clients, 1.2 is a sensible default in the absence of version in the received data.

Just removed it. The detail above is contained in https://github.com/w3c/rdf-concepts/pull/280

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/pull/150.html" title="Last updated on Jul 23, 2026, 4:44 PM UTC (64753c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-turtle/150/a84ef51...64753c4.html" title="Last updated on Jul 23, 2026, 4:44 PM UTC (64753c4)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -422,7 +422,7 @@
 
     <p>
      In the absence of a current version directive, any version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered.
-      In any case, the version announcement is merely a hint;
+      In any case, the version announcement is a hint;
       parsers are not required to reject features that are outside the announced version
       (but could signal them with a warning).
     </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -411,9 +411,9 @@
       This allows parsers that do not support these features to detect
       the presense of such features early, and potentially 
       inform the user, giving them an opportunity to stop the job
-      or otherwise act on the fact that the input data 
-      will not be entirely processed.
-    </p>
+      or otherwise act on the fact that the output data 
+      may be a subset of the intended output.
+</p>
 
     <p>Multiple version directives may appear in a Turtle document.
     The directive applies to the part of the document following the directive,
--- a/spec/index.html
+++ b/spec/index.html
@@ -409,7 +409,7 @@
       or <a href="#grammar-production-version"><code>@version</code></a>
       directives early in the document.
       This allows parsers that do not support these features to detect
-      the presense of such features early, and potentially 
+      their possible presence early on, and potentially 
       inform the user, giving them an opportunity to stop the job
       or otherwise act on the fact that the output data 
       may be a subset of the intended output.
--- a/spec/index.html
+++ b/spec/index.html
@@ -411,8 +411,8 @@
       This allows parsers that do not support these features to detect
       the presense of such features early, and potentially 
       inform the user, giving them an opportunity to stop the job
-      or otherwise act on the fact that some amount of the input data 
-      will not be processed as desired.
+      or otherwise act on the fact that the input data 
+      will not be entirely processed.
     </p>
 
     <p>Multiple version directives may appear in a Turtle document.
--- a/spec/index.html
+++ b/spec/index.html
@@ -412,7 +412,7 @@
       their possible presence early on, and potentially 
       inform the user, giving them an opportunity to stop the job
       or otherwise act on the fact that the output data 
-      may be a subset of the intended output.
+      may be a subset of the anticipated output.
 </p>
 
     <p>Multiple version directives may appear in a Turtle document.
--- a/spec/index.html
+++ b/spec/index.html
@@ -422,7 +422,6 @@
 
     <p>
      In the absence of a current version directive, any version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered.
-     In the absence of any version information, a sensible default is `"1.1"`, to maximize compatibility with the previous version of Turtle [[TURTLE]].
       In any case, the version announcement is merely a hint;
       parsers are not required to reject features that are outside the announced version
       (but could signal them with a warning).


spec/index.html
```diff
@@ -422,7 +422,6 @@ <h3>Version Announcement</h3>
 
     <p>
      In the absence of a current version directive, any version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered.
-     In the absence of any version information, a sensible default is `"1.1"`, to maximize compatibility with the previous version of Turtle [[TURTLE]].
       In any case, the version announcement is merely a hint;
```
afs:
```suggestion
      In any case, the version announcement is a hint;
```

spec/index.html
```diff
@@ -411,8 +411,8 @@ <h3>Version Announcement</h3>
       This allows parsers that do not support these features to detect
       the presense of such features early, and potentially 
       inform the user, giving them an opportunity to stop the job
```
afs, line 413: This reads to me as if the version is a commitment that features will be present.

If an RDF 1.2 feature is encountered, it is a parse error - this is stated in #148.

(Wording is made complicated for 1.2 because `VERSION` is one such feature!)

spec/index.html
```diff
@@ -411,8 +411,8 @@ <h3>Version Announcement</h3>
...
-      or otherwise act on the fact that some amount of the input data 
-      will not be processed as desired.
+      or otherwise act on the fact that the input data 
+      will not be entirely processed.
```
afs: "will not be entirely processed" - seems to say that the correct action is to skip.

TallTed:
```suggestion
      will not lead to the desired output.
```

pchampin: this change has the same issue as the original wording, IMO: it sounds like the unrecognized features could result *in something*, but that something would be different from the intended result -- in other words, it would be **misinterpreted**. The point we want to make is that this never happens. Each triple is either interpreted correctly, or not interpreted at all.

spec/index.html
pchampin: Well, it is true that the use of `VERSION` is discourage if non RDF 1.2 features are used, precisely because it could needlessly break old parsers. But this is allowed, just like using 1.2 features without VERSION-announcing them is also allowed.

spec/index.html
```diff
@@ -411,8 +411,8 @@ <h3>Version Announcement</h3>
       This allows parsers that do not support these features to detect
       the presense of such features early, and potentially
```
afs:
```suggestion
      the presence of such features early, and potentially 
```

spec/index.html
afs: At the triple level, yes.

But ingesting partial data is invalid at the domain level - e.g. data validated SHACL is no longer value.
Can we say something about failing to interpret a feature means that the data is subset (i.e. partial) which is more definite than "not entirely procesed".

pchampin: I agree that the *output* of the parser will be "a subset"/"partial" (and I agree that, at the application domain, that may render it invalid).

But we are talking about the *input data* here (i.e. the turtle source), so for me "not entirely processed" works best.

spec/index.html
```diff
@@ -422,7 +422,7 @@ <h3>Version Announcement</h3>
 
     <p>
      In the absence of a current version directive, any version specified as part of the <a href="#sec-mediaReg">Media Type</a> is considered.
-      In any case, the version announcement is merely a hint;
+      In any case, the version announcement is a hint;
```
TallTed, line 425:
```suggestion
      In any case, the version announcement is not a mandate, but a hint;
```

afs: I don't think this adds anything.

It does pose the question "what's a mandate?" - a new piece of terminology.

spec/index.html
afs: "that the output is subset"

I thought the idea was to not encourage fixup with RDF terms that weren't intended e.g. strings.

spec/index.html
```diff
@@ -411,8 +411,8 @@ <h3>Version Announcement</h3>
       This allows parsers that do not support these features to detect
       the presense of such features early, and potentially 
       inform the user, giving them an opportunity to stop the job
-      or otherwise act on the fact that some amount of the input data 
-      will not be processed as desired.
+      or otherwise act on the fact that the input data 
+      will not be entirely processed.
     </p>
```
pchampin:
```suggestion
      or otherwise act on the fact that the output data 
      may be a subset of the intended output.
</p>
```

spec/index.html
pchampin: See alternate proposal below (on line R414-R416)

spec/index.html
pchampin: I agree with Andy; I think the rest of the sentence ("parsers are not required") explains this enough.

spec/index.html
pchampin: alternatively (and to respond to @afs 's comment below that "it reads as if the version is a commitment"):
```suggestion
      their possible presence early on, and potentially 
```

spec/index.html
afs: Works for me.

afs (approved):

domel (approved):

ktk:
@pchampin please merge (discussed today)

spec/index.html
```diff
@@ -409,11 +409,11 @@ <h3>Version Announcement</h3>
...
       directives early in the document.
       This allows parsers that do not support these features to detect
-      the presense of such features early, and potentially 
+      their possible presence early on, and potentially 
       inform the user, giving them an opportunity to stop the job
-      or otherwise act on the fact that some amount of the input data 
-      will not be processed as desired.
-    </p>
+      or otherwise act on the fact that the output data 
+      may be a subset of the intended output.
```
TallTed: I think this is a more accurate description.
```suggestion
      may be a subset of the anticipated output.
```

